PR #5, opened 2019-05-03
Changes to genres and organised clones

Ryvius86:
Changed many game genres
Organised Sonic clones
Other minor tweaks


Ryvius86:
If a file is uploaded before a pull has finished it gets added automatically. I think that's what tripped me up.

PR #7, opened 2019-05-05
libretro backport & cleanup

barbudreadmon:
- backporting a few changes to make building easier on some platforms (mainly ps3 and android iirc)
- the version of the libretro port included here has been broken for years, let's remove it

@dinkc64 Make sure none of this is breaking standalone before merging ! Also, the latest version of the libretro port adds a ton of new files (not including generated files), do you prefer i keep that stuff on the libretro fork or should i commit them here too ?


dinkc64:
I'll try it after merging, lazy sunday :)
Also, what kinda files will be added?  can I see?

barbudreadmon:
That would be the cyclone cpu core and the content of https://github.com/libretro/fbalpha/tree/master/src/burner/libretro

dinkc64:
It might be best to leave them separate, for now.

PR #11, opened 2019-05-11
Load Game menu update

Ryvius86:
- Switched Title and Preview
- Formatted various text throughout
- Rearranged Filter tree now functions as intended and now has sub-categories
- Added a couple more rom hacks and updated another
- Other minor tweaks

Check here for a more in-depth list of changes: https://neo-source.com/index.php?topic=3472.0

Really glad to have gotten this finished.


dinkc64:
very nice!  there are a few minor things that need to be changed though, I'll change them after the merge though. (hw filter needs to stay open, since we use this a lot)

dinkc64:
I found that the list gets desynched from the checked state if you do something like:
1: deselect everything
2: select sega master system only
3: select/unselect the sega group
it's even worse with capcom, but, I think it can be easily fixed.  let me take a peek..

dinkc64:
I fixed the desync of checks, (latest check-in).  the config variables for cap&sega groups aren't needed since nLoadMenusomethingX takes care of it all.

PR #13, opened 2019-05-12
Remove links to the bad place

tmaul:
Also, I grabbed the translation files so we can host them in a way that is not controlled by Koch Media.


tmaul:
I added an index.html 

Possibly this could be hosted directly at github so then people can do pull requests on updated translations, etc.

dinkc64:
thanks Kev!!
Doing that was on my todo list, but for some reason I forgot about it, d'oh! :)

PR #14, opened 2019-05-14
Fix 64-bit mingw build

superctr:
Due to a race condition, `m64make.exe` did not run causing make to complain about missing `m68kops.c`. I did not check the other makefiles but the condition might be present there too.

Make invocation `make mingw510 -j5 BUILD_X64_EXE=1`
Make version `GNU Make 4.2.1`


dinkc64:
This is present in 32bit as well with -jXX - thank you for the fix :)

PR #15, opened 2019-05-14
Accurate QSound emulation

superctr:
Replaced the old sound emulation with the accurate one (written by me) from VGMPlay/MAME.

This is HLE so CPU usage should not be very high; the same emulator used in VGMPlay can render a complete song in a few seconds.

Some remaining issues so far:
- Not 100% tested.
- ssf2 has byteswapped samples, need to update ROM definitions to fix this.
- dino song 17 in the sound test has clicking, this is probably related to either ROM loading or timing differences (latter most likely).
- This code does not use Hungarian notation unlike the rest of FBA source code.

### DIFF
--- a/src/burn/drv/capcom/cps.h
+++ b/src/burn/drv/capcom/cps.h
@@ -329,6 +329,7 @@ void QscExit();
 INT32 QscScan(INT32 nAction);
 void QscNewFrame();
 void QscWrite(INT32 a, INT32 d);
+UINT8 QscRead();
 INT32 QscUpdate(INT32 nEnd);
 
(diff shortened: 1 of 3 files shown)


barbudreadmon:
Amazing, i see your project is GPL though so it might be safer to ask : you don't mind having your code re-licensed for non-commercial usage here ?

dinkc64:
This is amazing, thank you superctr!!

superctr:
This code was originally released under the BSD-3-clause (as it is in the MAME source code), but as I wrote this code I'm ok with licensing under the FBAlpha license.

PR #18, opened 2019-05-15
All I wanted to do was clean up the todo against hcastle...

tmaul:
I had a look to see why hcastle was crashing when using the correct refresh rate. It turns out that nBurnSoundLen is not calculated correctly until after DrvInit if a game is not using 60hz. This probably affects a lot of games and possibly might need further verifcation on others to make sure that the sound is working correctly now it's calculated correctly. Some sound cores didn't care about nBurnSoundLen when creating their internal buffers but that ones that do would always have a buffer underrun.

I also think I have found a fix for starwars crashing randonmly on my machine but if someone could verify that the games still work on other machines that would be good ;)

Some notes:
Wingforc - someone should check that the changes I made are correct :)
hcastle - seems ok with the refresh rate now
yiear - might be ok as is now I have renabled the refrest rate. Might not be. Could someone check.
namcos1 - Seems the games that are above 60hz now work with the correct refresh but someone should double check

Remaining refrest problems I know of:
metlclsh - Had a note that it locks up with refrest rate, seems to still crash so I have left it uncommented
M62 - Wasn't sure if the refrest rate was commented out due to sound so have left this

### DIFF
--- a/src/burn/drv/pst90s/d_kaneko16.cpp
+++ b/src/burn/drv/pst90s/d_kaneko16.cpp
@@ -4743,7 +4743,9 @@ static INT32 BlazeonInit()
 static INT32 WingforcInit()
 {
 	INT32 nRet = 0, nLen;
-
+	
+	BurnSetRefreshRate(59.1854); // hmm, this causes clicks in audio. let's just give it the cycles/per frame it wants instead. (see WingforcFrame())
+	
 	Kaneko16NumSprites = 0x4000;
 	Kaneko16NumTiles = 0x4000;
 	Kaneko16NumTiles2 = 0;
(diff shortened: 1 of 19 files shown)


dinkc64:
Hi Kev,
I would like to fix the older soundcores so that refreshrate can be called anywhere in init without needing it in a certain location of the init.  What I had been doing with new soundcores is allocate a 0x4000 buffer, which basically covers everything.

src/burn/drv/pst90s/d_kaneko16.cpp
```diff
@@ -7214,8 +7214,11 @@ static INT32 WingforcFrame()
 	Kaneko16MakeInputs();
 
 	INT32 nInterleave = 256;
-	nCyclesTotal[0] = ((UINT64)16000000 * (UINT64)10000) / 591854;
-	nCyclesTotal[1] = ((UINT64)4000000 * (UINT64)10000) / 591854;
+	nCyclesTotal[0] = 12000000 / 59.1854;
```
dinkc64: How do you get 12mhz here?  :)

tmaul: Copy and paste error, I think lol.

PR #26, opened 2019-05-22
This clause doesn't really make sense

barbudreadmon:
Since the project is reworking its license, it might be a good opportunity to remove this clause, it's pretty vague and i don't think that's something enforcable. If it is read verbatim it might even prevent the project from being usable at all (the project relies on directx, which is a bunch of dlls).

### DIFF
--- a/src/license.txt
+++ b/src/license.txt
@@ -5,7 +5,6 @@ You may freely use, modify, and distribute both the FB Neo source code and binar
  - You must include, verbatim, the full text of this license;
  - You may not distribute binaries which support games with copyright dates less then three years old;
  - You may not distribute FB Neo with ROM images unless you have the legal right to distribute them;
- - You may use dynamic linked libraries (dlls) for network play only (all other use is prohibited);
  - You may not ask for donations to support your work on any project that uses the FB Neo source code.
 
 FB Neo can currently be obtained from https://neo-source.com.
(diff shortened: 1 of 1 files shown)


src/license.txt
```diff
@@ -5,7 +5,6 @@ You may freely use, modify, and distribute both the FB Neo source code and binar
  - You must include, verbatim, the full text of this license;
  - You may not distribute binaries which support games with copyright dates less then three years old;
  - You may not distribute FB Neo with ROM images unless you have the legal right to distribute them;
- - You may use dynamic linked libraries (dlls) for network play only (all other use is prohibited);
  - You may not ask for donations to support your work on any project that uses the FB Neo source code.
```
dinkc64: I'm with you there, but 2 things:
what does Kev, iq_132 think?
what do you guys think they meant when that line was written?  
At first I thought, maybe someone would use an external dll to shovel in ads (remember the shitware crisis of the early 00's?)  Who knows... One can only guess...

barbudreadmon: Yeah, that's why i requested their review. I think Kev already agreed on the forum though. I've absolutely no idea why this line was included, i guess only barry might know.

tmaul: I really have no idea at all. I originally thought it might have come from MAMEs old licence but I can't see anything in that either. I guess it raises the question can we change the licence without permission of all the licence holders, considering the way it reads means that every version of finalburn from now since the start has broken that part of the licence by loading in DLLs (for Direct X to use the above example)?

barbudreadmon: Honestly, i'm not even sure we can change the name of the license, but we are forced to do it anyway, so we might as well remove anything we consider unrealistic/unenforcable/unintelligible.

barry obviously trampled his own rights by throwing his own license into the bin so he is not a matter of concern. We might also ask @zxmarcos and @vbt1 who contributed to the project if they are ok with this change. If you have anyone else in mind, let's tag them.

tmaul: For what it's worth, I vote remove it as the emulator can't exist without DLLs.

dinkc64: barbudreadmon it really doesn't concern them.
tmaul, alright :)

PR #38, opened 2019-05-28
SNES driver "improvements"

tmaul:
I've opened this as a pull request as I am still unable to build a windows build with GCC on my machine so it's only been tested in visual studio so I didn't want to just commit without someone being aware :)

Things that are in this:

* Initial attempt at sound (although it's not working well, so watch your ears lol)
* 1st pass at refactoring of the CPU. It fails every accuracy test rom I can throw at it so I thought if I at least make it easier to read someone could actually have a chance of finding and fixing the issues. Might be worth just porting a different CPU core. Probably this also needs to be done to the other snes files as well.
* general attempt at clean up but there is still much more to do unfortunately. 

The final aim with this is not get get all the snes games added to FBNeo (imagine the rom scan wait time!) but at least get it to the point where we can add a curated list of the best ones that run ok.


dinkc64:
If you can someday make this play Super Ghouls and Ghosts with sound, you'll be my hero for a day or 2 :)

dinkc64:
did I say day, I meant month or 2 :)

PR #50, opened 2019-06-09
Sync with MAME (romset names)

barbudreadmon:
The names of those romsets don't match with MAME, most likely because MAME added them after us and our names didn't follow their naming policy (?). Any reason for not syncing names here ?


dinkc64:
usually when we originate something, It's preferred that we with our names/ways.  BUT, considering who actually named them (an ex developer who we will call "the robot"), I think it might be a good idea to move on.

PR #59, opened 2019-06-23
cps1: drawing sf2 at vblank (fix logo zoom/rotation)

barbudreadmon:


dinkc64:
wait - see discussion in bug report :)

PR #66, opened 2019-06-30
fix mame link

igdegoo:


igdegoo:
> 
> 
> Wouldn't be better set as https instead?

done

tmaul:
Thanks!

PR #127, opened 2019-10-10
Raspberry Pi build

0xe1f:
Allows building emulator on Raspberry Pi. Uses SDL for sound and some of the core features, native libraries for video and input.

Where possible, I added/updated common code in SDL directories, in case someone decides to continue maintaining the SDL-only build.


0xe1f:
I've tested it on 3, but it should run on all but 4. I'll update the source once I figure out how to deal with 4's cryptic video renderer.

0xe1f:
Thanks! You can commit now - it builds and runs. I'll submit more PRs for additional work (like Pi4 support).

tmaul:
Works really well here, might be time to dust off the ol' picade.

PR #131, opened 2019-10-13
SDL build

0xe1f:
I've updated the stale SDL makefile to include all the recent additions to drivers, etc. I tested on Linux (works fine) and mac OS (issue with window focus, but works).

Building on Linux should work out of the box, however, building on macOS requires `gcc-9` (`brew install gcc`) - 4.2.1, which is included with Xcode segfaults while building the Musashi core.

I've copied `dynhuff.h/.c` to `dep/dynhuff/`, since it's referenced outside win32-specific code. I didn't want to move references in win32-specific code from `win32/dynhuff.h/.c`, since I don't have access to VS to verify that the code will build, but it would probably be a good idea to do it, so that there aren't two copies floating around.

I've made a few other updates, including some affecting the Pi build.


tmaul:
@0xe1f I have moved dynhuff to the main burner directory. Can you rebase and check to make sure it's all good like this for the SDL build please?

0xe1f:
Done, thanks

tmaul:
@0xe1f Cool! also feel free to update the whatsnew with the SDL updates and rpi port if you wish :)

0xe1f:
Sure, is that whatsnew.html? Should I update notes for v0.2.97.44 notes, or add a new one?

tmaul:
Yes, that's the one, and please stick them in the v0.2.97.44 bit and I will merge this over. Then when I feel less crappy I will see if I can update appveyor to include the SDL build, but that will take a while :)

0xe1f:
Feel better soon!

PR #159, opened 2019-10-30
macOS build update

0xe1f:
This commit adds mouse support, cursor locking, fixes DIP switch support and other misc fixes.


0xe1f:
This last commit changes where generated files are written (now in the `DerivedSources` directory designated by Xcode).

It also fixes a header naming bug in the sprite/tile generating scripts. The bug manifests itself when part of the output path (not the file) contains a period, causing the output path for the `_table` headers to get mangled; e.g.:

`<snip>/Intermediates.noindex/<snip>/cave_tile_func.h`
should generate a table file named
`<snip>/Intermediates.noindex/<snip>/cave_tile_func_table.h`
but instead, it attempts to generate a file named
`<snip>/Intermediates_table.noindex/<snip>/cave_tile_func.h`
, because the regex is anchored against the first period. This commit fixes that, causing the last period to get substituted instead.

@tmaul I'm almost certain that this should fix the compilation problem you're experiencing.

0xe1f:
.. although now it's barking that it can't find `SDL` :) https://github.com/0xe1f/FBNeo/runs/281072924#step:3:0
We'll get there.

tmaul:
That sounds like some progress at least :)

PR #165, opened 2019-11-06
Types sanitize

barbudreadmon:
I was thinking about doing those changes, is there any reason we shouldn't ?
1st commit fixes some declaration that would prevent 2nd commit to build.


barbudreadmon:
Need proper testing with savestates, don't merge for now

Issue #173, opened 2019-11-19
pi core hangs with missing roms


grant2258:
There is probably more  glamorous ways ways to do this and get the error messages. In this case its just to close the emulator on a rom load failure.  adding this to the pi main() should do the job. 

        if (DrvInit(driverId, 0) !=0)
        {
                printf("driver init failed\n");
                return 1;
        }

tmaul:
This also happened with the generic SDL version, which I just commited this fix for. However, I don't have a pi with me to test that build so I have not fixed this on the pi version :)

grant2258:
it does work on the pi version just want to see how it reacts to bad crc and nodumps if thats an issue might need to add options for defined behavior if it errors. Thats the only things i can see causing any possible issues. Im just not sure what the default behavior is atm still getting familiar with the code base.

0xe1f:
The next pull request will fix this in the pi version - thanks!

grant2258:
nice to see that fixed job well done. One other problem i have is i cant seem to get my controllers to work on the pi the keyboard is fine. I made the joyconfig/default.jc 
```
BUTTON1=1
BUTTON2=2
BUTTON3=3
BUTTON4=4
BUTTON5=5
BUTTON6=6
COIN=8
START=9
```
one of the buttons should have done something ill look more into it at the weekend when I have more time so will close this for now

grant2258:
I took the sdl driver code and added some code got the joysticks working on my barcade. I will tidy it up a bit during the weekend if anyone's interested in getting the joysticks working on sdl

0xe1f:
Joysticks were working in the SDL build - what was the issue?

grant2258:
Havent tried the sdl build yet only the pi one I just borrowed the sdl code for that and added the joystick mapping instead of default keyboard for it. 

Ill give the sdl build a shot  havent tested that one yet only ill get back to you on that one. The pi joysticks arent working all as it is its just in the readme the mac version says joysticks arent working so just assumed that was the case it used sdl.

https://github.com/finalburnneo/FBNeo/blob/master/README.md

grant2258:
just add this to line 321 to test the sdl code and how its mapping
``` 
int nJoyNumber = (nCode - 0x4000) >> 8;
printf("(joystick) joy%d  ncode:%x state:%d mapped:%s\n", nJoyNumber, nCode, JoystickState(nJoyNumber, nCode & 0xFF), InputCodeDesc(nCode));
```
[...]

PR #174, opened 2019-11-20
Pi 4 Support

0xe1f:
This PR adds initial support for rendering via DRM, so the emulator should now be playable on the Pi 4. It's likely unoptimized and could use additional tweaking; I'll be chipping away at it as time allows.
It also properly handles driver init errors by propagating them up the initialization chain (#173).


tmaul:
Cool :)

PR #175, opened 2019-11-22
Added basic documentation & renderer info

0xe1f:
Small modification that includes instructions on how to build on Pi 4 and renderer info for the DRM build (for troubleshooting).

It's clear that FBNeo isn't alone in rendering issues on Pi4 - internet is awash with complaints about poor fps performance with code that otherwise worked fine on earlier versions of Pi.  It's possible that this is largely due to vast majority of code being based on the same rendering example (some rendition of https://github.com/eyelash/tutorials/blob/master/drm-gbm.c), but it's also possible that the drivers used by the new Pi still need work.

I will run some additional profiling to see if I can pin down the source of the problems, but this is probably it for a while - as far as Pi 4.

### DIFF
--- a/README-PI.md
+++ b/README-PI.md
@@ -1,12 +1,25 @@
 # FinalBurn Neo for Raspberry Pi
 
 ## Compiling
-To compile the emulator, run `make pi`. You will need SDL libraries for sound
-and input.
+To compile the emulator, run `make pi`. You will need SDL 1.2 libraries for
+sound and input:
+
+```
+sudo apt-get install libsdl1.2-dev
+```
+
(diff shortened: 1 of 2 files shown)


0xe1f:
Some additional notes that I also posted on Twitter:
1. Performance can be considerably better with a decent power adapter
2. Rendering starts very smooth after a reboot, and degrades somewhat rapidly from there.  Degradation persists across processes, so there's likely a resource leak somewhere
3. I've tried caching framebuffers post-swap - this doesn't seem to help (if anything, it seems to have a negative effect)

PR #176, opened 2019-11-23
update sdl input

grant2258:
This is a update for the sdl port controls it allows the user to set the command line with -joy. 

Without -joy player1 will be keyboard and player2 will be joy0 if its available. 
With -joy  player1 will be joy0 and player2 will be joy1 this should help people with barcade setups to get up and running.

SF2 Mapping needs implemented if there is any intensest in it being done is doable.

### DIFF
--- a/src/burner/sdl/main.cpp
+++ b/src/burner/sdl/main.cpp
@@ -9,28 +9,61 @@ There are lots of problems with the audio output code.
 There are lots of problems with the opengl renderer
 probably many other things.
 ------------------*/
+
+
 #include "burner.h"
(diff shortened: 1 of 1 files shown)


grant2258:
@0xe1f I been trying this on another version of linux  that has different numbering than stretch for the xbox controller. I think it varies over platform. 

I think a workable solution might be to print text on the screen on first run to configure controllers like you do on retropie on initial setup (only the buttons) so they can be mapped like they should be. Im going to close this one and rethink a solution. SDL seems to be more sane with the joystick mapping but a solution can be found for sdl1

PR #199, opened 2019-12-25
update dat creation

grant2258:
We cant use the all in one system dat because of name conflicts so make all the dats that are available.

### DIFF
--- a/src/burner/dat.cpp
+++ b/src/burner/dat.cpp
@@ -272,7 +272,7 @@ INT32 write_datfile(INT32 bType, FILE* fDat)
 		if (BurnDrvGetTextA(DRV_SAMPLENAME)) {
 			strcpy(ssName, BurnDrvGetTextA(DRV_SAMPLENAME));
 		}
-
+#if !defined(BUILD_SDL2)
 		remove_driver_leader(HARDWARE_SEGA_MEGADRIVE, 3, 1)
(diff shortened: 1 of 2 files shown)


grant2258:
Just a little update I fixed the global datfile creation for sdl2 only. I know libretro uses the subsystem to do this. so the for example the aburner rom had a conflict with the arcade rom. When i took away the lead removal it lets us call the game at the command prompt with ./fbneo spec_aburner. Im not sure why the lead was removed but im sure there is good reasons just to let you know its only the sdl2 port this will change on perhaps the sdl1 port as well.

grant2258:
Seems this is good for building the global dat will need to dig a little more to launch the roms with the names so dont pull yet till. The gui is listing the roms with these names as well must be changing it somewhere else as well.

tmaul:
Yeah, the prefixes are filtered out, they are just there so it doesn't screw up the internal gamelist. They see filtered out when loading, with the assumption that each system has its own room folder to avoid .zip name clashes

grant2258:
its not really a big deal to be honest can use different folder. I like everything in one directory might just make this updated for myself will see what i end up with and you guys can make the choice if you want it or not for the sdl build for now ill just change this so it creates all the dats with no functionality changes.

grant2258:
Ok changed it to just create the dats for all the systems.  for both sdl versions the global isint usable at the moment

grant2258:
closing this one and opening a new one.

PR #203, opened 2019-12-26
alt dat option

grant2258:
Ok here I have an optional way to make a dat and it will have no effect unless you define ALT_DAT  in the makefile. This effectively removes the zip name conflicts. The zips can be put in one directory or separated the same way as they used to be then only difference is it will use the driver names instead of trimming them.

@tmaul please note you will need to rescan i was scratching my head why is wasnt working for me. If you guys do not want this its no problem at all. I just like my roms in one dat file its easier keeping them up to date. This will also help working on fixing the remaining conflicts in the dats

### DIFF
--- a/src/intf/video/sdl/vid_sdl2.cpp
+++ b/src/intf/video/sdl/vid_sdl2.cpp
@@ -101,12 +101,12 @@ static int Init()
 	}
 
 	sdlWindow = SDL_CreateWindow(
-		title,                  // window title
-		SDL_WINDOWPOS_CENTERED, // initial x position
-		SDL_WINDOWPOS_CENTERED, // initial y position
-		display_w*nScale,              // width, in pixels
-		display_h*nScale,              // height, in pixels
-		screenFlags             // flags - see below
+		title,
+		SDL_WINDOWPOS_CENTERED,
+		SDL_WINDOWPOS_CENTERED,
+		display_w*nScale,
+		display_h*nScale,
+		screenFlags
 	);
 
 	// Check that the window was successfully created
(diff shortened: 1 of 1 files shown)


tmaul:
@grant2258  I have merged in the main changes, but the defs for the main alt dat define I will let @dinkc64  sign off and then I will merge those in as well. It seems like a good idea to me fwiw

dinkc64:
@grant2258,
Is there another route you can take rather than editing all of the drivers using GetZipName?

dinkc64:
also, do you rename all of your roms to sms_gamename, md_gamename etc?  if so, you're doing it wrong :)  we have things in place that takes care of all of this in the win32 version, as long as each system is in it's own directory.

grant2258:
Yea this basically keeps the driver name. I can see the name of the system is being trimmed off. I dont really mind if this isint for the project. I just want to give some perspective on the reasoning behind it and offering it if you wanted to use it. 

One selfish reason is more for my personal use the way i store my roms in one directory(I know not the best reason). 

SDL2  and windows does have a gui so is wont be an issue there is a practical reason for it if your using the sdl port with no menu and you look at your roms and goto the spectrum directory and you see aburner.zip so you go to command prompt and type ./fbneo aburner and the arcade version starts you would need to know in advance what name to put in it. I know separating into directories will solve this with a gui.

The only other point of entry without changing the drivers would be in the getzipname function and re-add the trimmed leads there with the ALT_DAT define(I though it would be more practical in the driver though).

Again its more for my personal use if it has no place here  no need to add it at all. I do have concerns about this in that all romsets should work on all versions of fba and i understand it from that perspective. Its a simple change i can do for myself on my personal build just thought I would share the options if its a viable one.

grant2258:
messed this pr up with a merge will close if its need in the future its an option we can look at

PR #205, opened 2019-12-27
Incorrect commenting fixed

Loggan08:


src/dep/libs/lib7z/7zTypes.h
```diff
@@ -4,10 +4,6 @@
 #ifndef __7Z_TYPES_H
 #define __7Z_TYPES_H
 
-#ifdef __LIBRETRO__
-#include "streams/file_stream_transforms.h"
-#endif
-
```
dinkc64: @barbudreadmon, is this ok?

src/dep/libs/lib7z/7zFile.h
```diff
@@ -4,7 +4,7 @@
 #ifndef __7Z_FILE_H
 #define __7Z_FILE_H
 
-#if defined(_WIN32) && !defined(__LIBRETRO__)
+#ifdef _WIN32
```
dinkc64: @barbudreadmon, what about this?
@loggan08, what are the reasons for removing the libretro if's (here and below in 7zTypes.h)?

barbudreadmon:
Most likely @Loggan08 overwrited the lib7z library without looking if there were hacks in the previous one.
Furthermore, i looked at this lib7z upgrade, it seems utterly pointless ? The only changes, besides removing my libretro hacks, are about bumping dates and moving/removing comments around...

src/dep/libs/lib7z/7zTypes.h
Loggan08: I grabbed the SDK from LZMA's website, unless you do not directly support the libretro-port then it should be ok, otherwise I will revert those changes.

src/dep/libs/lib7z/7zFile.h
Loggan08: > Most likely @Loggan08 overwrited the lib7z library without looking if there were hacks in the previous one.
> Furthermore, i looked at this lib7z upgrade, it seems utterly pointless ? The only changes, besides removing my libretro hacks, are about bumping dates and moving/removing comments around...

It's the same SDK version, i tried bumping it, but it requires a rewrite of un7z.cpp

Loggan08:
A tip from my experience with code, if enabling SSE2 for builds, it should provide a minor boost to performance.

Intel Pentium 4 supports SSE2

Loggan08:
For un7z.cpp , the CLookToRead, (if upgrading to 19.00) has to be changed to CLookToRead2.

Tried that, but there were still some inconsistencies needing fix.

src/dep/libs/lib7z/7zTypes.h
barbudreadmon: should be ok now

Loggan08:
yuh, 7z changes are working after compiling (and loading 7z roms)

barbudreadmon:
I'm worried those `stddef.h` to `cstddef` changes are gonna break support for some odd compilers...

Loggan08:
> I'm worried those `stddef.h` to `cstddef` changes are gonna break support for some odd compilers...

It's in the scope of C++ language, any ideas on which compilers ?
Let the appveyor first check out whether the GCC build is successful.

barbudreadmon:
> It's in the scope of C++ language, any ideas on which compilers ?

i was thinking about the console SDKs, i know the ps3 one has an odd implementation.

src/burn/drv/irem/d_m92.cpp
```diff
@@ -9,7 +9,7 @@
 #include "msm6295.h" // ppan
 #include "irem_cpu.h"
 #include "iremga20.h"
-#include "stddef.h"
+#include <cstddef>
```
dinkc64: no, please don't change that - it's not compatible with super ancient compilers that some systems need to use :)

0xe1f:
These changes aren't even "comments" - they're #defines and #includes?

Issue #132, opened 2019-10-13
Weird extra scanline in VS builds


tmaul:
As noticed by @dinkc64 

There seems to be something odd about the game area (placeholder screen / emulation area) being pushed down 1 line of pixels.
This happens in both the X64 and 32 VS builds, not the gcc builds.
It happens regardless of blitter.  Here's a pic of what I mean:

VS 32/64bit screencap 
![fb-vs](https://user-images.githubusercontent.com/50554883/66721748-9b541b00-edfd-11e9-9f8e-62a2bcd9fbda.png)

GCC build:

![fb-normal](https://user-images.githubusercontent.com/50554883/66721750-a27b2900-edfd-11e9-9eee-01eda81919bf.png)

tmaul:
I thought this might be a .rc file thing where maybe VS2019 is compiling it differently than GCC but I have no idea if that's even a thing.

barbudreadmon:
bad math ? feels like some kind of rounding issue to me

tmaul:
Some more notes: This is only happening with certain blitters. E.g. doesn't happen with softfx. As @dinkc64  predicted it is happening in int ScrnSize()

tmaul:
Also, at least for me, this is happening in GCC builds with the experimental and dx9 alt blitters.

dinkc64:
That's weird!  I can't make this happen here :/
Even with the latest win32-gcc.7z from the download link.

mobasuite:
maybe it's been solved since the past two months

tmaul:
@dinkc64  That's very interesting. Is there any chance you can send me a copy of your own build via the forum or something? 

I now think its one of a few things:
1) graphics drivers
2) a bug in VS and appveyor GCC builds (but then that would mean that everyone would see it)
3) still a bug somewhere in the insides of burner

dinkc64:
Sure :)  sent!

tmaul:
This still happens with the build from @dinkc64 

I'm still looking at it when I have time :)

dinkc64:
That's really strange, I remember when it was happening and I could see it - but now it won't reproduce on my pc with any of the builds.  gremlins!

mobasuite:
It's your monitor.

dinkc64:
Bug finally fixed, big thanks to tmaul, he gets at least half the credit here, without his support, I wouldn't have been able to run VS on my pc to do anything :)

best regards,
- dink

PR #207, opened 2019-12-27
Summary of previous PR

Loggan08:
This is for the second commit :
Capture


src/burn/snd/asteroids.cpp
```diff
@@ -20,7 +20,7 @@
  *
  *****************************************************************************/
```
dinkc64: I'm afraid we can't support this as it will lock out all of the other systems which use fbn.  We have to stay backwards compatible.  sorry about that!

src/intf/interface.cpp
```diff
@@ -8,26 +8,26 @@ INT32 IntInfoFree(InterfaceInfo* pInfo)
 		for (INT32 i = 0; i < INT_INFO_STRINGS; i++) {
 			if (pInfo->ppszInterfaceSettings[i]) {
 				free(pInfo->ppszInterfaceSettings[i]);
-				pInfo->ppszInterfaceSettings[i] = NULL;
+				pInfo->ppszInterfaceSettings[i] = nullptr;
 			}
```
dinkc64: we also can't support nullptr due to the same reason (backwards compiler compatibility).  It's very important that fb stays compatible with ancient hw :)

src/burn/snd/asteroids.cpp
Loggan08: You mean xbox/xbox360 ?

Again, those changes are in the scope of C++ language, any way to confirm this ?

src/intf/interface.cpp
Loggan08: I'm 99% positive it does improve the experience on Windows OS

tmaul: Is it possible that it's fixed that vs2019 extra line issue #132?

grant2258: you could make a global variable for the make file "ancient_cpp" and define it for platforms that need it else it uses the more optimized route for later platforms nothing wrong with supporting both.

Loggan08: Capture

src/burn/snd/asteroids.cpp
dinkc64: I know for a fact it won't compile on the compiler iq_132 uses, so this can't happen

dinkc64: we have to stay compatible with c++03, nullptr is c++11

src/intf/interface.cpp
dinkc64: we're not going to be changing the c++standard or modernizing anything, sorry about that.

In what way does changing NULL to nullptr improve the experience in windows?  I think that blank line in vs is coming from a bug in scrnsize in burner/win32/scrn.cpp

Loggan08: @dinkc64 

I can revert the nullptr changes, if you wish.

src/burn/snd/asteroids.cpp
Loggan08: will revert the nullptr stuff

PR #208, opened 2019-12-28
Summary of PR (2)

Loggan08:
reverted some nullptr stuff


Loggan08:
- Only the interface has been touched, drivers and such remain untouched (with the Null -> nullptr)

PR #211, opened 2019-12-29
Test before merge

Loggan08:
Might require some testing before 'ok'


grant2258:
there is still some c11 includes in the main fbneo code

Loggan08:
> there is still some c11 includes in the main fbneo code

what will c11 break ?

grant2258:
only saying dink mentioned it in the other post we need to stick to the older standard for older compilers on some ports

barbudreadmon:
@Loggan08 we won't merge this, we already told you we want to be compatible with older compiler, and we especially don't want c11 to be required.

Loggan08:
Does'nt this fix something : 

https://github.com/finalburnneo/FBNeo/pull/211/commits/b676ac8c702cdfc4e484097a014a7c005e3f7f24#diff-88c667869b5299ec26b675124c5dbf64

Loggan08:
- Summary of changes :

Increase un7z-speed.
Fix : https://github.com/finalburnneo/FBNeo/issues/132
Random performance enhancments (at least on WinOS)

Loggan08:
There's no reason not to do a cpp-language update, for legacy compilers/systems one can have a driver_old.cpp

barbudreadmon:
> There's no reason not to do a cpp-language update, for legacy compilers/systems one can have a driver_old.cpp

There is no reason to do a cpp-language update, we told you we don't want that, we need backward compatibility, period. We don't plan on cloning every drivers (and the work maintaining them !) either.

Closing this PR, and i hope i won't have to repeat we don't want c11 again.

PR #215, opened 2019-12-31
Windows-OS Enhancments

Loggan08:
- libpng updated to git version
- SSE2 enabled for x86 msvc builds
- versionhelpers api for detecting WinVer


src/burner/win32/misc_win32.cpp
```diff
@@ -1,29 +1,18 @@
 #include "burner.h"
-
+#include <VersionHelpers.h>
 bool bEnableHighResTimer = true; // default value
```
dinkc64: Please keep in mind:
versionhelpers.h is for msvc (aka VS20xx) only, most of us use mingw/gcc.

When working on source files, try to respect that file's layout & format:
for example, burner/win32/sel.cpp uses char* p;
but other files might use char *p;  No need to change this, the author of the file wanted it this way.

stricmp should remain stricmp (same with the wcs* stuff)

Loggan08: will edit today

Loggan08: edited, let me know if there is anything else.

PR #222, opened 2020-01-03
Needs review

Loggan08:


dinkc64:
fba -> fbn is nice, but 
Code restructures are generally not accepted (everything else)

dinkc64:
"If it ain't broke, don't fix it", especially if it has been working fine for years.  Old people used to tell me this when I was a kid, they were right. :)

Loggan08:
I don't see unsafe changes with this one, which code restructures ?

PR #226, opened 2020-01-05
cpp language update

mobasuite:
at least compiles on msvc x64


barbudreadmon:
We don't want a "cpp language update", fbneo needs to be backward compatible with old compilers, closing.

barbudreadmon:
you seem to be the same person as loggan08, so i'm gonna make this clear : we don't want any of your PR, at this point you are just spamming, which is very annoying, thanks for your understanding.

PR #236, opened 2020-01-11
Mapping

grant2258:
@tmaul can you test this out your end. Ill test this properly tomorrow when i get more time dont pull it yet still need to add sf2 mapping should get that done tomorrow you should be able to use your controller to coin up now.


tmaul:
Cool, I will check it out. My linux PC has developed a bit of problem so just reinstalling that before I can test.

grant2258:
Tested fine with me here as long as sdl2 has the right mapping for your controller you should be good to go.

tmaul:
@grant2258 This seems to work pretty well here, but only tested with a couple of joypads. Shall I merge it in now or are you still adding to it?

grant2258:
passed with the pads i have as well we can fix anything if any issues pop up

PR #243, opened 2020-01-21
(Super Puzzle Fighter II Turbo) Combined compatibility with Boss/Color hacks

ghost:
These are currently compatible with one hack or the other, but mixing them isn't compatible, this commit is to fix this.

### DIFF
--- a/d_cps2.cpp
+++ b/d_cps2.cpp
@@ -1,0 +1,30 @@
+STD_ROM_PICK(Spf2tcbh)
+STD_ROM_FN(Spf2tcbh)
+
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Maybe i'm overthinking this but i'm kinda against it, there will be no end to the number of clones if we start combining hacks into new romsets, furthermore i think the last time someone added a mix of hacks, the author(s) of the hacks complained.

barbudreadmon:
Well, account deleted & the code was actually wrong, i'll close.

dinkc64:
whats the deal with these hit-n-run pr's lately??

PR #250, opened 2020-01-25
Fix sound and refresh rate of some Nichibutsu games

sasuke-arcade:
- Fix refresh rate for some Nichibutsu games (see MAME)
- Fix DAC Sound pitch and gain of Armed Formation hardware, especially the pitch of orchestra hit of Armed Formation

### DIFF
--- a/src/burn/drv/pre90s/d_armedf.cpp
+++ b/src/burn/drv/pre90s/d_armedf.cpp
@@ -77,7 +77,6 @@ static INT32 usemcu = 0;
 static INT32 Terrafjb = 0;
 static INT32 Kozuremode = 0;
 static INT32 Skyrobo = 0;
-static INT32 fiftysevenhertz = 0;
 
 static struct BurnInputInfo ArmedfInputList[] = {
 	{"P1 Coin",		BIT_DIGITAL,	DrvJoy1 + 10,	"p1 coin"	},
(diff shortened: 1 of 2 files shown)


src/burn/drv/pre90s/d_armedf.cpp
```diff
@@ -822,7 +821,7 @@ static INT32 DrvSynchroniseStream(INT32 nSoundRate)
 
 static INT32 DrvSyncDAC()
 {
-	return (INT32)(float)(nBurnSoundLen * (ZetTotalCycles() / (6000000.000 / (nBurnFPS / 100.000))));
+	return (INT32)(float)(nBurnSoundLen * (ZetTotalCycles() / (((6000.0 / nBurnFPS) * 6000000.000) / (nBurnFPS / 100.000))));
 }
```
dinkc64: Just pointing out: either way returns exactly the same calculation :)

sasuke-arcade: Thanks! This code should have been revert.
Armed F refresh rate is 57 before the change, so pitch of DAC sound was wrong. This code was to fix it.
But later I noticed that the refresh rate was wrong and fixed it to 59.08 (same as MAME).
It seems that the DAC pitch was corrected only by changing refresh rate, and this code was needless.

PR #260, opened 2020-01-26
g_armedf.cpp crash and minor fixes

sasuke-arcade:
- Revert unneeded change in PR #250.
- Fixed crash on player change in Armed Formation two-player game. This was caused by overflow memory access in sprite drawing.

### DIFF
--- a/src/burn/drv/pre90s/d_armedf.cpp
+++ b/src/burn/drv/pre90s/d_armedf.cpp
@@ -1248,7 +1248,7 @@ static void draw_sprites(INT32 priority)
 		INT32 clut  = spr[offs + 2] & 0x7f;
 		INT32 sx    = spr[offs + 3];
 		INT32 sy    = sprite_offy + 240 - (attr & 0x1ff);
-		code       &= 0xfff;
+		code       &= 0x7ff;
 
 		if (*flipscreen) {
 			sx = 320 - sx + 176;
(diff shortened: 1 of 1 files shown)


src/burn/drv/pre90s/d_armedf.cpp
```diff
@@ -1260,7 +1260,7 @@ static void draw_sprites(INT32 priority)
 		sy -= yoffset;
 		sx -= xoffset;
 
-		if (sx < -15 || sy < -15 || sx >= nScreenWidth || sy >= nScreenHeight) continue;
+		if (code > 0x7ff || sx < -15 || sy < -15 || sx >= nScreenWidth || sy >= nScreenHeight) continue;
```
dinkc64: Hi,
Instead of checking code like that, how about fix the bitmask:
		code       &= 0xfff;
change to 
		code       &= 0x7ff;

sometimes upper bits of sprite code is used for internal game stuff, masking it will give the correct code.

best regards,
- dink

sasuke-arcade: I checked the source of latest MAME and confirmed that only changing the mask to 0x7ff was fine.
I committed the change it.
Thanks for review.

dinkc64: welcome :) Looks good!

PR #264, opened 2020-01-28
d_terracre.cpp : Fixed sprite lag

sasuke-arcade:
Kid no Hore Hore Daisakusen, Soldier Girl Amazon, Terra Cresta
- Fixed one frame lag in sprite and BG moving. Modified in the same way as the latest MAME.


dinkc64:
Thanks :)

PR #267, opened 2020-02-01
armedf, galivan, terracre, snk, ssrj, markham : Some fixed

sasuke-arcade:
g_armedf.cpp
- Tuned DAC volume based on comparisons to PCB sound
- Deleted the volume of each game since they all had the same volume after the tuning

g_galivan.cpp
- Tuned DAC volume based on comparisons to PCB sound
- Fixed wrong DIPSW assignment

g_terracte.cpp
- Changed default DIPSW from cooktail to upright

g_snk.cpp
- Fixed Athena, Ikari, Victory Road music tempo

g_ssrj.cpp
- Add start button assignment (Same fix as MAME 0.216)

g_markham.cpp
- Fix BanBam / Pettan Pyuu sound pitch


dinkc64:
thanks :)

PR #274, opened 2020-02-09
Neo Geo Secret Hacks

WandeingLagomorph793:
With some help from Dinkc64, everything should be set up properly, this adds in compatibility with a handful of secret character hacks for a few Neo Geo games.


dinkc64:
Would you mind if I checked this in from my side?  For some strange reason it things you wrote the entire d_neogeo.cpp file again. :)

WandeingLagomorph793:
> 
> 
> Would you mind if I checked this in from my side? For some strange reason it things you wrote the entire d_neogeo.cpp file again. :)

Sure thing! I tried uploading it as originally zipped (after extraction) but it would fail the checks when I did that.

barbudreadmon:
the path is wrong, it should be `src/burn/drv/neogeo/d_neogeo.cpp` but it's only `d_neogeo.cpp` instead

PR #276, opened 2020-02-09
Neo Geo Hacks

WandeingLagomorph793:
With the edited path now, errors out when I'd try editing manually.

This is for a small handful of character hacks for Super Dodgeball, Magical Drop III, the two Real Bout Fatal Furies, and Waku Waku 7, more will come later but I have to set those up first.


dinkc64:
It worked out quite good!   Thanks for sticking around and learning some things:)

WandeingLagomorph793:
> 
> 
> It worked out quite good! Thanks for sticking around and learning some things:)

No problem! :)
Soon I'm gonna try setting up some more of these, I still got that link with a few more, it'll take awhile since it's a matter of checking the crc checksums, but eventually I'll be able to mark it down. Good that I got the learning experience from this too, that'll certainly help! :)

PR #277, opened 2020-02-10
remove snes driver, and also the makefile for the old MAME build setu…

tmaul:
…p as it doesn't work with the current one


tmaul:
@dinkc64  Do you think this is a good idea? The snes driver is never going to go anywhere without a complete re-write, and the mamemingw makefile doesn't work with the current mame toolchain, and never really worked that well anyway.

dinkc64:
Personally I think its a good idea, writing a snes emulator that works perfectly with most games is an undertaking that would be about 3-4x more insane than the nes project.

dinkc64:
I missed the second part - I think we should leave in mamemingw stuff.  I'm pretty sure this is how JacKc compiles his binaries.

barbudreadmon:
With special chips & all, it would be an insane project indeed, if the current code won't lead us anywhere, i'm not against dropping it.

PR #281, opened 2020-02-19
Patches for wofh & wofha

SumavisionQ5:
to fix the boss not coming out on 1st stage.
Patches are ripped from RAM with running FBAS.


dinkc64:
Thank you :)

SumavisionQ5:
> Thank you :)

Glad to be helpful :)

PR #285, opened 2020-02-22
Terra Force (d_armedf.cpp) : Fixed blinking screen in name entry and service mode

sasuke-arcade:


dinkc64:
very nice, thanks :)

PR #294, opened 2020-03-03
outrun: support for hiscores

barbudreadmon:
It turns out it won't work for every game, for `toutrun` it's saving something but fail applying it : 
```
Applied Hi Score Memory Range 0 on frame number 35583
Failed attempt to apply Hi Score Memory Range 1 on frame number 35583
```

`outrun` hiscores work properly though, any idea ?


XjmllPZ:
Thank you for reaching out ❤️

barbudreadmon:
@dinkc64 i confirm toutrun hiscores are working properly without the second line in hiscore.dat

barbudreadmon:
The issue seems to be with the hiscore.dat file itself (`@:maincpu,program,60909,1,09,09` must be removed), so let's merge...

PR #303, opened 2020-03-09
skyarmy: Fix bitrot, makes the game fully playable [O. Galibert]

arcadez2003:
New dump of the main CPU to fix game breaking problems in level 3


arcadez2003:
Might be worth waiting a little while until the new rom is widely available.??

PR #304, opened 2020-03-09
sharrier: Better protection simulation [Adrian Smethurst]

arcadez2003:


arcadez2003:
Now that the bad man is gone i think i can touch the SEGA code without someone
getting all butt hurt :)

PR #312, opened 2020-03-24
FBN already supports the redumped rom

arcadez2003:
0.157: DsNo redumped/verified Fight Fever romsets. Added new 060-s1.s1 rom.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -12882,12 +12882,7 @@ struct BurnDriver BurnDrvtwsoc96 = {
 static struct BurnRomInfo fightfevRomDesc[] = {
 	{ "060-p1.p1",    0x100000, 0x2a104b50, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 
 
-#if !defined ROM_VERIFY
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Does it work right ? The previous exception was obviously intentional since there was the ROM_VERIFY stuff.

arcadez2003:
Well i tested the change on the xbox FBN port and the game works fine, however the neogeo
driver in FBL is over 9 years old now and does not contain that ROM_VERIFY code.

I simply swapped out the the older 060-s1.s1 rom and replaced it with the new dump i see
no reason why the changes i made would not work with the the current FBN builds.

// Fight Fever (set 1)

static struct BurnRomInfo fightfevRomDesc[] = {
	{ "060-p1.p1",    0x100000, 0x2a104b50, 1 | BRF_ESS | BRF_PRG }, //  0 68K code

//	{ "060-s1.s1",    0x020000, 0x7f012104, 2 | BRF_GRA },           //  2 Text layer tiles

    { "060-s1.s1",    0x020000, 0xd62a72e9, 2 | BRF_GRA },           //  2 Text layer tiles

barbudreadmon:
Ok, fine by me, we can still fix any issue that may arise later

PR #319, opened 2020-03-31
Add new redump of b04401w064.u1 for the Do-Donpachi 3 sets

arcadez2003:
b04401w064.u1 on all the other ddpdoj sets was found to be corrupt (doesn't match checksum on ROM) this was redumped and replaced in the sets, exept for in the bootleg set, which is based around the corrupt dump (nw)

Credits Shoutime - David Haywood


arcadez2003:
I've not added support for the new ddpdoj japan location test version which was added to MAME
at the same time as the redumped b04401w064.u1. it requires twice as many graphical roms and
i need more experience with this core before attempting to update PGMGetRoms to accomodate
double the amount of gfx roms.

PR #323, opened 2020-04-10
Redumped gulunpa ROM 1

arcadez2003:
The game is still marked as non working in MAME even with the new dump of the bad gfx rom, although this could simply be an oversight i'll leave it as not working here for the moment.


dinkc64:
thanks!

XjmllPZ:
thanks!

PR #339, opened 2020-04-26
Fix incorrect sizes for mslug5b files

Sanaki:
The sizes of ms5b-c5.bin and ms5b-c6.bin were initially misreported. I've personally verified the files otherwise.


Sanaki:
Fixed a few minutes before I submitted this PR.

PR #360, opened 2020-05-22
Mamemingw changes

tmaul:
Are these changes ok? 

Basically I have changed the mamemingw makefile so it doesn't require mingw mkdir, which means FBNeo can now be built with http://www.winlibs.com/ out of the box, which is nice if you want an easy way of using GCC10 for example. I have checked with @JacKc029735 and he very kindly confirmed that this doesn't break his build environment. 

Building with GCC10 threw up a warning in inps.cpp. The change I made seemed like the most sensible fix, and as far as I can tell doesn't break any other build when running it and trying that dialog. I have no done any win32 programming for ages though so it's possible I have made a terrible mistake...

Finally, I added a BUILD_NATIVE flag to all the GCC makefiles, and if this is defined it will build using -march=native -mtune=native. This would be a terrible thing to enable for releases, but is handy if you are building your own binaries as it will try and make GCC enable all the cpu specific optimisations for the machine you are building on, as I know a lot more people are doing that nowadays.

### DIFF
--- a/makefile
+++ b/makefile
@@ -24,6 +24,9 @@ UNICODE = 1
 # Include x86 Assembly routines
 BUILD_X86_ASM = 1
 
+# Include GCC optmisations for your CPU e.g use -march=native. WARNING: This might mean that the generated binaries will not run on other peoples (older) machines!
+#BUILD_NATIVE = 1
+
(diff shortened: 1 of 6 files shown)


dinkc64:
Looks good to me :)  no need to request a review, if something breaks we'll fix it!

tmaul:
Cool :)

PR #362, opened 2020-05-24
add support for ngp dat

barbudreadmon:
Either i'm really bad at this, or having `HARDWARE_SNK_NGP | HARDWARE_SNK_NGPC` for the bios is causing way too many issues with the dat file generation, so let's only keep `HARDWARE_SNK_NGPC` ?


barbudreadmon:
btw, do not merge !

dinkc64:
this can be removed now

PR #365, opened 2020-05-26
gaplus : Starfield fix and add service mode in DISPW

sasuke-arcade:
- Fixed displaying of starfield
  - Fixed scrolling speed and direction ([MAME Testers-00434](https://mametesters.org/view.php?id=434))
  - Added top and bottom clipping ([MAME Testers-07663](https://mametesters.org/view.php?id=7663))
  - Modified to use sprite color instead of text color with refer to the PCB
  - Added blinking when the movement of the star changes
- Add service mode in DIPSW setting


dinkc64:
Very nice, thanks for your contribution! :)

PR #409, opened 2020-07-28
Bigrun graphics OK for big endian machines

crystalct:


dinkc64:
nice work, thank you!

PR #410, opened 2020-08-01
pgm: big-endianize attempt 2

barbudreadmon:
@dinkc64 re big-endianizing those handlers : finally i wonder if it's not required, 2 reasons : 
- it fixes the following error on ddp2
ddp2-200801-051400
- there is something similar done in m68k : 
https://github.com/finalburnneo/FBNeo/blob/b51b55f2bdef3f21dbffd501984f857841a151ea/src/cpu/m68000_intf.cpp#L333
https://github.com/finalburnneo/FBNeo/blob/b51b55f2bdef3f21dbffd501984f857841a151ea/src/cpu/m68000_intf.cpp#L375-L378

### DIFF
--- a/src/burn/drv/pgm/pgm_run.cpp
+++ b/src/burn/drv/pgm/pgm_run.cpp
@@ -408,7 +408,7 @@ static UINT16 __fastcall PgmZ80ReadWord(UINT32 sekAddress)
 	pgmSynchroniseZ80(0);
 
 	sekAddress &= 0xffff;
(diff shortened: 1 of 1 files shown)


barbudreadmon:
On a sidenote, without this change, i generally hear some barely audible clics when inserting coins, and some more clics that somehow sounds like the music on ddp3.
I also noticed some games will give me a black screen when loaded after another, example : svgpcb after svg.

barbudreadmon:
it also turns out i get sound by reverting https://github.com/finalburnneo/FBNeo/commit/8b1689a83cb318965ad92dcb136dc063a47cde8f , so it seems something introduced in this commit requires big-endianization

PR #418, opened 2020-08-07
Big endian first attempt K053250 - Bishi now works

crystalct:


barbudreadmon:
i might be mistaken, but the fixes on k056832 look a bit weird to me, with the way it's declared i wouldn't expect `k056832Regs` to need all those endian fixes.

crystalct:
When a 16bit matrix is used sometimes as matrix to store 16bit UINT16 variables, somethimes as memory to manage UINT16 pointer and somethimes as memory to manage UINT8 pointer, after 3 days of attemps, i solved endiainied all references to it.

dinkc64:
Nice work, are you finished with this set or do you plan to add more to it?

crystalct:
Dgz, xexex, mysticwar and others aren't full working.... Konami and big endian still not full friends

crystalct:
Gijoe, for example, sprites aren't visible. But K053247.cpp was endianized in the past and other games that use it are working.

barbudreadmon:
gijoe could be using codepaths in k053247 that aren't used by any other games

crystalct:
How is possible that gijoe_sprite_callback is never colled? Is it a sprite collision callback?

dinkc64:
gijoe_sprite_callback is called for every single sprite.

barbudreadmon:
if it's never called, i guess the conditions increasing count in K053247SpritesRender are never true.

crystalct:
```
	offs = count = 0;

	if (zcode == -1)
	{
		for (; offs<0x800; offs+=8) {
			if (BURN_ENDIAN_SWAP_INT16(SprRam[offs]) & 0x8000) {
				sortedlist[count++] = offs;
			}
		}
	}
	else
	{
		for (; offs<0x800; offs+=8)
			if ((BURN_ENDIAN_SWAP_INT16(SprRam[offs]) & 0x8000) && ((BURN_ENDIAN_SWAP_INT16(SprRam[offs]) & 0xff) != zcode)) {
				sortedlist[count++] = offs;
			}
	}
```

There, zcode is -1 but `sortedlist[count++] = offs;` nevel called (if always false)
count remain 0 and then sprice callback isnt called.
Same code called from xmen works.

barbudreadmon:
Assuming this code is right because it works for xmen, it means SprRam is filled the wrong way for gijoe. It looks to me like either gijoe_objdma isn't called the way it should, or is writing crap to K053247Ram when it's called. Imho, log every calls to that function and compare them with x86.

barbudreadmon:
imho, conditions for `gijoe_objdma` being called aren't met, meaning either `K053246_is_IRQ_enabled` or `K056832IsIrqEnabled` is returning crap

crystalct:
And this:

```
void K053246Write(INT32 offset, INT32 data)
{
	if (data & 0x10000) { // handle it as a word 
		*((UINT16*)(K053246Regs + (offset & 6))) = BURN_ENDIAN_SWAP_INT16(data);
		
	} else {
		K053246Regs[offset & 7] = data;
	}
}
```
Its meaning is obscure for me.... and on gijoe `if (data & 0x10000)` is never true.

barbudreadmon:
Seems legit to me : if word swap, if byte don't swap

crystalct:
Yes, right....  K053246Regs is UINT8

crystalct:
this:
```
static void __fastcall gijoe_main_write_word(UINT32 address, UINT16 data)
{
	if ((address & 0xfffff8) == 0x110000) {
		K053246Write((address & 0x06) + 0, data >> 8);
		K053246Write((address & 0x06) + 1, data&0xff);
		return;
	}

```
Is it only one little endian way to write a word to an UINT8 matrix?

barbudreadmon:
Maybe try
```
#ifdef LSB_FIRST
		K053246Write((address & 0x06) + 0, data >> 8);
		K053246Write((address & 0x06) + 1, data&0xff);
#else
		K053246Write((address & 0x06), data);
#endif
```

dinkc64:
try this instead:
#ifdef LSB_FIRST
			memcpy(dst_head, src_head, 0x10);
#else
			for (INT32 i = 0; i < 0x10; i++) {
				dst_head[i] = BURN_ENDIAN_SWAP_INT16(src_head[i]);
			}
#endif

barbudreadmon:
@crystalct i looked into it, `gijoe_objdma` was called, so the issue was about what it was writing to sprite ram, it turns out the condition `if (*src_head & 0x8000)` must be turned into `if (BURN_ENDIAN_SWAP_INT16(*src_head) & 0x8000)`

crystalct:
i'll try

dinkc64:
Nice work friends, keep it up :)

crystalct:
Work!!!!
image

crystalct:
Commit if you want.
Mystic Warrior still blank screen....Maybe it isn't easy to fix....

dinkc64:
again, nice work :) thank you!

barbudreadmon:
@crystalct i'm sure you'll figure out mystic warrior, when you are really stuck try to print and compare all kinds of variable with x86.
btw i won't be home next week so libretro syncing will probably be delayed.

crystalct:
I'll Do It ^_^
If you find a way to sync libretro i will make (finally) FB Neo RL plus
v2.1 (PS3) next week.

Il dom 9 ago 2020, 12:35 barbudreadmon <notifications@github.com> ha
scritto:

crystalct:
If mystwarr_tile_callback is never called and K056832Init works in others games.... where could be the problem?

dinkc64:
since this one was committed, please open a new pr and we will try to figure it out :)

PR #422, opened 2020-08-14
Viostorm big endian OK

crystalct:


dinkc64:
Nice fix :)  Now I realize why it needs to be on the byte write - the registers array is UINT16, but it's UINT8 pointing to them.  (instead of the other usual way: UINT16 pointing to UINT8..) :)

best regards,
- dink

PR #456, opened 2020-09-20
nes: nes_shadonin fix clone

taoenwen:


barbudreadmon:
this one is part of the robot-generated entries, so no merging, let's wait for @dinkc64, i also think he'll make this the other way around (US as parent and JP as clone)

taoenwen:
@barbudreadmon It's a great idea

dinkc64:
thanks, "robot" is updated and fix checked in :)

PR #469, opened 2020-09-27
macro & autofire: Fix for sdl2

taoenwen:
Solve the problem that Windows API causes compilation errors on sdl2


Sanaki (approved):
Confirmed this fixes the SDL2 build. Thanks for a quick fix!

barbudreadmon:
Thanks, it looks fine on libretro linux too.

PR #480, opened 2020-10-02
Add switchres to linux/sdl2

substring:
This allows perfect display on CRT
The work is based on https://github.com/antonioginer/switchres which is the switchres code extracted from GroovyMAME and made available to other emulators that wish to add CRT display.
This first commit simply concerns linux using the SDL2 video backend.
Add INCLUDE_SWITCHRES=1 to enable switchres functionnality at make.
To test under Linux you need:
  * a valid switchres.ini in /etc
  * libswitchres.so in your LD_LIBRARY_PATH

The sdl2 target may work under Windows, not tested.

Current limitations :
  * if switchres couldn't set the right refresh rate, nothing is made for emulation to properly work with the real screen refresh_rate
  * missing in-game resolutions change
  * need to improve screen rotation for vertical games

What is required to make this work once compiled :
  * libswitchres.so from the switchres repo mentionned above
  * a switchres.ini in /etc (there is a sample in the switchres repo once again)


tmaul (commented):
Note for whoever: please leave this pr open for testing until substring is happy :)

Note for sdl2 users: please test this, it's very cool.

substring:
Removed switchres_wrapper.h, improved vertical games.

Switchres now has a `make install` tat makes everything necessary available system wide.

What's cruelly missing for SDL2:
  * screen rotation detection (if the screen is vertical, we're doomed)
  * adapt emulation to the screen refresh rate. Technically speaking, vertical games hardly have any chance to run at their real refresh rate on a horizontal screen. We get switchres' screen refresh rate, emulation should eventually be adapted to it.

substring:
Updated the readme. Limitations updated to:

  * the SDL2 backend doesn't support rom interlace. Although switchres can compute an interlace resolution, FBNeo isn't yet passing that flag.
  * Vertical screens are not detected, so vertical games will always be rendered as horizontal
  * in-game resolution switching is not yet supported
  * if the monitor's refresh rate is too different from the game refresh rate, no mechanism comes to smooth the video output, so tearing can happen.

PR #514, opened 2020-10-21
Neo Geo Hacks

WandeingLagomorph793:
Neo Geo hacks for Fatal Fury, Art of Fighting, and King of Fighters, alongside assorted games like the first Samurai Shodown and Matrimeelee, mostly to allow the boss characters without the need of a button combination.

I believe I have the errors figured out now, looking at the build I spotted redefinition errors, so I went back and cleaned those up, tried to get the formatting and everything correct as well.

### DIFF
--- a/d_neogeo.cpp
+++ b/d_neogeo.cpp
@@ -1,0 +1,20431 @@
+// FB Alpha Neo Geo driver module
+
+#include "neogeo.h"
+#include "bitswap.h"
+
+#if !defined ROM_VERIFY
+	// enable to match Neo Geo BIOS to MAME's (also affects dipswitches -- leaves only the MVS BIOS option)
+	// #define ROM_VERIFY
+#endif
+
+static struct BurnRomInfo emptyRomDesc[] = {
+	{ "",                    0,          0, 0 },
+};
+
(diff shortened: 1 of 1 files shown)


dinkc64:
Hi, 
appveyor says there is a problem compiling:
d_neogeo.cpp:20087:1: error: expected unqualified-id before '/' token
 / Fatal Fury 2 / Garou Densetsu 2 - arata-naru tatakai (Boss Hack)
 ^

also, can you re-create the PR?  the first file makes github think you've authored the entire d_neogeo.cpp

thanks,
- dink

WandeingLagomorph793:
> Hi,
> appveyor says there is a problem compiling:
> d_neogeo.cpp:20087:1: error: expected unqualified-id before '/' token
> / Fatal Fury 2 / Garou Densetsu 2 - arata-naru tatakai (Boss Hack)
> ^
> 
> also, can you re-create the PR? the first file makes github think you've authored the entire d_neogeo.cpp
> 
> thanks,
> 
>     * dink

I'll give it a shot. Not sure what it is with Fatal Fury 2's that's giving the issue. I'm gonna try a PR without FF2 specifically to see if that one works.

PR #515, opened 2020-10-22
Neo Geo Additional Hacks

WandeingLagomorph793:
Adjusted some formatting and took out Fatal Fury 2's hack in case that was causing errors.

### DIFF
--- a/d_neogeo.cpp
+++ b/d_neogeo.cpp
@@ -1,0 +1,20431 @@
+// FB Alpha Neo Geo driver module
+
+#include "neogeo.h"
+#include "bitswap.h"
+
+#if !defined ROM_VERIFY
+	// enable to match Neo Geo BIOS to MAME's (also affects dipswitches -- leaves only the MVS BIOS option)
+	// #define ROM_VERIFY
+#endif
+
+static struct BurnRomInfo emptyRomDesc[] = {
+	{ "",                    0,          0, 0 },
+};
+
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Please don't send PR anymore, or change method to do them ('add files via upload" is obviously a crappy method), all of the PR you send have the same issue.

WandeingLagomorph793:
Alright. I recall doing these exact steps last time and they worked so I'm still trying to adjust and figure out what's going wrong. It looked like I was getting close on this one, I'm trying to make sure it's formatted as it should while looking back at what worked before, maybe it's just how it's being uploaded.

barbudreadmon:
Maybe, tbh i only use command-line git so i don't know how "adding files via upload" works

WandeingLagomorph793:
I've just tried doing that and noticed the additions numbering at +574 rather than the +20,000 it was the last few times, I'm thinking it was trying to put the d_neogeo.cpp file in the main folder directory, not the /src/burn/drv/neogeo folder that it's supposed to go into, so it might have had build issues with that I'm thinking.

 I just tried on my own with actions enabled and it seemed to build correctly when I tried the new method, like manually going into the folder and going from there, though I'm uncertain of any discrepancies between those actions compared to how the Appveyor builds it. 

I also double checked the formatting to make sure I didn't miss something in the file itself that could break it, comparing these to the ones from last time like Real Bout 2.

PR #520, opened 2020-10-23
Neo Geo Hack Set

WandeingLagomorph793:
Adjusted the NULL, HACK and CLONE tags for the appropriate new hacks. not sure if I need to *replace* any of those parent sections, like p1.p1 or if keeping it there should be fine for the hack, I'm thinking that's part of what causes Samurai Shodown's checksum error so I'm trying with this for now. AOF3 and KOF95 don't replace any files for their respective hacks.

Tried my luck with the tabbing sections for a good hour so far and haven't had any luck with this

	{ "069-p1.p1",    0x100000, 0xa8bcfbbc, 1 | BRF_ESS | BRF_PRG }, //  0 68K code			/ TC538200
    { "069-p1bh.p1",    0x100000, 0xB8362F59, 1 | BRF_ESS | BRF_PRG }, //  1 68K code			/ TC538200
	{ "069-sp2.sp2",  0x200000, 0xdbe963ed, 1 | BRF_ESS | BRF_PRG }, //  2 					/ TC5316200

I tab into the messed up section in the middle and it isn't fixing for some reason, I backspace and tab and for some reason it's just not sticking, I tried fixing it via github but tabbing into it from there only made the issue worse. I'm having a ton of trouble with this one. I think it's the last thing that I need to fix but I've been stuck for awhile trying to fix it.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -18633,263 +18633,240 @@ struct BurnDriver BurnDrvbrkrevext = {
 	0x1000,	320, 224, 4, 3
 };
 
-// Garou - Mark of the Wolves (Boss Hack)
-/* Original Version - Encrypted GFX */ /* MVS VERSION - later revision */
+// Garou - Mark of the Wolves (Enable hidden characters)
 
-static struct BurnRomInfo garoubhRomDesc[] = {
+static struct BurnRomInfo garoubRomDesc[] = {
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Ok, there were just so many issues that i spent the last 2 hours replacing half of your code.
Next time you are adding hacks, **add them 1 at a time**, tbh if i tried to get you to fix your whole PR yourself, i think we would still have been at it 100 PRs later.

PR #540, opened 2020-11-02
SGX -  Homebrew conversion of RTA2 added

crystalct:
image


barbudreadmon:
thanks for your contribution, i didn't know there was an active sgx homebrew/hack scene, adding more games for that system is pretty cool

crystalct:
Unfortunatly there are very few project about SGX. I dont find any more, just other demos/tools from Chris Covell's Creations like:
Fractal Engine 2 and SuperGrafx Demo

PR #554, opened 2020-11-06
Some pst90s Big Endian fixes

crystalct:
Before continuing I ask you for an opinion about:
`static INT32 DrvScan(INT32 nAction, INT32 *pnMin)` and `*pnMin = 0x029707;`
In the past it never big endian swapped.... but i think it must be swapped.... and that is weird.
What is that value for?


barbudreadmon:
it should be save files versioning, and you started doing it in https://github.com/finalburnneo/FBNeo/commit/008498b601ce59af2c2c7c06a8d316d9ceeca9f7 but that shouldn't be necessary, except maybe if you intend to have cross-arch compatibility for save files (it would probably require a lot more work on the big endian fixes though).

dinkc64:
it's only to locally check the version number of the loaded state, please change these swaps back :)

crystalct:
Done.

PR #558, opened 2020-11-08
win32ui: Add a maximize and restore button to the form [select game]

taoenwen:
The gamelist window has a low initial height and limited visible content
1. Enlarge the window by dragging the edge of the window (fbneo's default scheme)
2. Add a button to maximize and restore the window (another way)


dinkc64:
taoenwen, it's very nice - thank you.

PR #573, opened 2020-11-13
d_kaneko16.cpp - Complete Big Endian fix

crystalct:


barbudreadmon:
@dinkc64 are you ok with those changes ?

dinkc64:
@crystalct, Can you do this without modifying kanekotb.h?

crystalct:
Who written (Luca Elia) original Kaneko16 driver needed to keep some instruction sets to be loaded into memory when needed.
For example:
```
/* MCU executed command: 0400 0E00 0034 - code*/
/*
200E00: 7071    moveq   #$71, D0
200E02: 7273    moveq   #$73, D1
200E04: 7475    moveq   #$75, D2
200E06: 7677    moveq   #$77, D3
200E08: 7879    moveq   #$79, D4
200E0A: 7A7B    moveq   #$7b, D5
200E0C: 7C7D    moveq   #$7d, D6
200E0E: 7E7F    moveq   #$7f, D7
200E10: 4E75    rts
*/
static const UINT16 bonkadv_mcu_4_34[] = {
	0x7071,0x7273,0x7475,0x7677,0x7879,0x7a7b,0x7c7d,0x7e7f,
	0x4e75
};
```
He used a natural way like to declare UINT16 arrays in source code and those arrays are used only for that (load bytes to memory). 
In that example, RTS instruction (4E75) in a 68000 (and in a x86) is stored in memory as 754E and a UINT16 store it exactly 754E (always in those machines). So in  Little Endian machines nothing else is needed other than `memcpy`.
Looking at mame's source they use a function to load bytes from those array and that function is different on Big endian machine rater than little endian, because it's necessary make a swap and copy inside a cycle.
FBNeo use the most efficient way as told before (`memcpy`) but it works only in little endian machines due to how UINT16 arrays are stored in memory. In Big Endian machine it should be made in a less efficent way like mame.
But why lose efficence and to discriminate Big Endian machines? Just declare those arrays as UINT8 and swap bytes direclty at source. memcpy just copy bytes, it need a source a destination and how many bytes to copy, so it's transparent about Little endian and Big Endian machine, infact the instructions to copy arrays to memory are exactly the same without needing to change them.
Having said that, if you prefer to keep the historical declaration of those arrays, i can make a differentiation inside kanekotb.h using 
```
#ifdef LSB_FIRST
			bla bla bla
#else
			bla bla bla
#endif
```
Or making a `kanekotb_bigendian.h` and make the differentiation of `#include` inside `d_kaneko16.cpp`.

dinkc64:
@crystalct, try this:
use original kanekotb.h, but replace like this in d_kaneko16.cpp:
#ifdef LSB_FIRST
#define MCU_RESPONSE(d) memcpy(&MCURam[MCUOffset], d, sizeof(d))
#else
#define MCU_RESPONSE(d) {						\
	INT32 rsize = sizeof(d);					\
	UINT8 *dest = (UINT8*)&MCURam[MCUOffset];   \
	UINT8 *src  = (UINT8*)&d;					\
	for (INT32 i = 0; i < rsize; i++) {			\
		dest[i] = src[i^1];						\
	}											\
}
#endif

dinkc64:
Sorry, please read this before the previous comment: :)
there is no performence to lose by using a for loop here, MCU_RESPONSE is called rarely: see: (number is frame number from boot)
mcu resp 0007
mcu resp 0007
mcu resp 0008
mcu resp 0008
mcu resp 0008 <- bootup
mcu resp 0965 <- begin of attract mode
mcu resp 2848 <- start game

crystalct:
Fixed your code a bit and done.

dinkc64:
Hi, sorry about the trouble, can you re-create the PR without changes to kanekotb.h?

@barbudreadmon, please don't merge anything that can't be diffed, for example - when a file says "too large to show/diff.."

barbudreadmon:
@dinkc64 actually there was no change to kanekotb.h in https://github.com/finalburnneo/FBNeo/commit/365d0f5e486eb9aa7234a6eca0233ce579899ce4 because i squashed commits

crystalct:
In my last commit i putted again original kanekotb.h, I thought that was enough. I'm sorry.

dinkc64:
oops, my bad, i made a mistake.  sorry!

crystalct:
image

dinkc64:
:)

PR #583, opened 2020-11-18
Update d_megadrive.cpp

Charles32k:
hacks letter L


dinkc64 (changes requested):
Charles32k,
please fix game-names (remove .bin from game name in game list)
also please use proper tab-indent like other games in the driver :)

thanks,
- dink

PR #584, opened 2020-11-19
Fatal Fury 1 Character Hack

WandeingLagomorph793:
Been awhile but another one of these. I've only got a few more and then that's all I got for Neo Geo, unless I stumble across a new character hack but the channel I found that covers these has most of those.

It was recommended last time that I start with just one of these to catch mistakes early, so that's what I'm doing, starting with Fatal Fury since that one was specifically requested. He also said who made this one, this specific hack was made by Yumeji, so it's in the title for that. After this next set of hacks I've got all the Neo Geo hacks I have, though there's a few from other systems, I got one for Asura Buster though I don't know which system that one goes into.

I'm seeing a few deletions and not entirely sure what's up with that one when I'm comparing this compared to the regular D_neogeo.cpp and I didn't seem to find where the deletions are coming from, It's the most recent version of that and for this PR I'm only putting in Fatal Fury so as far as I can tell it seems like there should just be additions, I looked through via comparing but didn't find where it was so if anybody knows where those are coming from I'll make sure to fix that.

I'm using the code editor Atom, alongside a browser based codesandbox to double check, I'm hoping to avoid any of the white space issues that seemed to crop up constantly last time. I also made sure it has the appropriate headers for clones, hacks, and so on.

I also found who made that mock PS1 hack for KOF97, I think, the one link even mentioning it's existence mentioned it being made by someone called EGCG/EGGT, I can edit that into it's entry but I'd rather not mess that up so I'd like to see how this goes first.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -145,7 +145,7 @@ static struct BurnDIPInfo ms5pcbDIPList[] = {
 	{0,		0xFE, 0,	2,	  "Stop mode"},
 	{0x00,	0x01, 0x80,	0x00, "Off"},
 	{0x00,	0x01, 0x80,	0x80, "On"},
-	
+
 	// Region
 	{0,		0xFD, 0,	2,	  "Region"},
 	{0x02,	0x02, 0x03,	0x02, "Asia / Europe"},
@@ -755,7 +755,7 @@ static struct BurnDIPInfo neogeoDIPList[] = {
 	{0x02,	0x01, 0x3f,	0x20, "Universe BIOS ver. 1.1"           },
 	{0x02,	0x01, 0x3f,	0x21, "Universe BIOS ver. 1.0"           },
 	{0x02,	0x01, 0x3f,	0x22, "NeoOpen BIOS v0.1 beta"           },
-	
+
 	// Memory card
 	{0,		0xFD, 0,	2,	  "Memory card"},
 	{0x02,	0x01, 0x80,	0x80, "Writable"},
@@ -861,7 +861,7 @@ static struct BurnDIPInfo neoFakeDIPList[] = {
 	{0x02,	0x01, 0x3f,	0x20, "Universe BIOS ver. 1.1"           },
 	{0x02,	0x01, 0x3f,	0x21, "Universe BIOS ver. 1.0"           },
 	{0x02,	0x01, 0x3f,	0x22, "NeoOpen BIOS v0.1 beta"           },
-	
+
 	// Memory card
 	{0,		0xFD, 0,	2,	  "Memory card"},
 	{0x02,	0x01, 0x80,	0x80, "Writable"},
@@ -1168,7 +1168,7 @@ static struct BurnRomInfo ms5pcbBIOSRomDesc[] = {
 	{ "",                  0x00000, 0x00000000, 0                                                   }, // 34
 	{ "",                  0x00000, 0x00000000, 0                                                   }, // 35
 	{ "sp-4x.sp1",         0x80000, 0xb4590283, BRF_ESS | BRF_PRG | BRF_BIOS                        }, // 36 PCB BIOS
-	
+
 	{ "",                  0x00000, 0x00000000, 0                                                   }, // 37 Z80 BIOS
 	{ "sfix.sfix",         0x20000, 0xc2ea0cfd, BRF_GRA | BRF_BIOS                                  }, // 38 Text layer tiles
 	{ "000-lo.lo",         0x20000, 0x5a86cff2, BRF_ESS | BRF_BIOS                                  }, // 39 Zoom table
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Wow, there were so many roaming spaces, thanks for the cleanup.

So, about the addition, i located the hack in hbmame : https://github.com/Robbbert/hbmame/blob/90c1e916dc407f5e03d1e6c97e059a80b03ef884/src/hbmame/drivers/ngrbff.cpp#L730
I usually prefer that we keep the same name and content than hbmame when the romset exists there, i believe it makes things easier for everyone, however in this case i don't like the s01 suffix so i'm fine if you want to keep the bh suffix instead. 


Now, analyzing the code from your PR : 

```
	{ "033-p1.p1",    0x080000, 0x47ebdc2f, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 		/ TC534200
	{ "033-p2.p2",    0x020000, 0xc473af1c, 1 | BRF_ESS | BRF_PRG }, //  1 					/ TC531024
	{ "033-p1bh.p1",    0x080000, 0xfc051e1d, 1 | BRF_ESS | BRF_PRG }, //  2 68K code 		/ TC534200
```
This is wrong, the `033-p1bh.p1` is a replacement for `033-p1.p1`, so the correct code should have been be : 
```
	{ "033-p1bh.p1",    0x080000, 0xfc051e1d, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 		/ TC534200
	{ "033-p2.p2",    0x020000, 0xc473af1c, 1 | BRF_ESS | BRF_PRG }, //  1 					/ TC531024
```
with the original `033-p1.p1` romset disappearing.


Next : 
```
	{ "033-m1.m1",    0x020000, 0x5be10ffd, 4 | BRF_ESS | BRF_PRG }, //  8 Z80 code 		/ TC531001
	{ "033-m1h.m1",    0x020000, 0xa8603979, 4 | BRF_ESS | BRF_PRG }, //  9 Z80 code 		/ TC531001
```
This one is very very wrong, at first i thought you just did the same mistake than the p1 rom, actually you did but that's not the biggest issue here, the rom with crc `0xa8603979` actually comes from another hbmame romset named `fatfury1d` : https://github.com/Robbbert/hbmame/blob/master/src/hbmame/drivers/mgd2.cpp#L451
I can't accept this for 2 reasons : 
1. we don't mix roms from different romsets
2. the header in hbmame says this one is a bad dump made with "Mega Game Doctor 2"

WandeingLagomorph793:
> Wow, there were so many roaming spaces, thanks for the cleanup.
> 
> So, about the addition, i located the hack in hbmame : https://github.com/Robbbert/hbmame/blob/90c1e916dc407f5e03d1e6c97e059a80b03ef884/src/hbmame/drivers/ngrbff.cpp#L730
> I usually prefer that we keep the same name and content than hbmame when the romset exists there, i believe it makes things easier for everyone, however in this case i don't like the s01 suffix so i'm fine if you want to keep the bh suffix instead.


That's good to know, The only reason I put the bh suffix is to try and avoid any future conflicts if there's an existing rom in the database that uses a more common suffix like just b for a regular bootleg set.

> 
> Now, analyzing the code from your PR :
> 
> ```
> 	{ "033-p1.p1",    0x080000, 0x47ebdc2f, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 		/ TC534200
> 	{ "033-p2.p2",    0x020000, 0xc473af1c, 1 | BRF_ESS | BRF_PRG }, //  1 					/ TC531024
> 	{ "033-p1bh.p1",    0x080000, 0xfc051e1d, 1 | BRF_ESS | BRF_PRG }, //  2 68K code 		/ TC534200
> ```
> 
> This is wrong, the `033-p1bh.p1` is a replacement for `033-p1.p1`, so the correct code should have been :
> 
> ```
> 	{ "033-p1bh.p1",  0x080000, 0xfc051e1d, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 		/ TC534200
> 	{ "033-p2.p2",    0x020000, 0xc473af1c, 1 | BRF_ESS | BRF_PRG }, //  1 					/ TC531024
> ```
> 
> with the original `033-p1.p1` romset disappearing. I think you may have misunderstood something last time when i said your hacks were missing parents, i was talking about this :
> 
> ```
> 	"fatfury1bh", "fatfury1", "neogeo", NULL, "1992",
> ```
> 
> In your last commit, "fatfury1" was replaced by NULL in lines like this one. Now it's good, and you didn't forget the BDF_CLONE flag either :)

Oh I see, I figured I needed the original parents in there or else it would throw up CRC errors, you mentioned that with Samurai Shodown I, which is one of the next ones I have, so I wanted to avoid that, but if it's not needed then I can clean this one up a little and submit a new PR if needed. Although one hack I have is for Samurai Shodown V, and that one has no matching files at all and in fact currently works if I put it's hack files in with an existing Samsho5 since it doesn't replace anything, but I don't know if HBMame supports that one since I haven't had the best luck running that, My current .cpp has all the files in place needed but that adds up to about 20 entries so I'm guessing all I would need is just files like "270-c3d.c3  0x82db9dae" and not that plus the normal 270-c3.c3"

> 
> Next :
> 
> ```
> 	{ "033-m1.m1",    0x020000, 0x5be10ffd, 4 | BRF_ESS | BRF_PRG }, //  8 Z80 code 		/ TC531001
> 	{ "033-m1h.m1",    0x020000, 0xa8603979, 4 | BRF_ESS | BRF_PRG }, //  9 Z80 code 		/ TC531001
> ```
> 
> This one is very very wrong, at first i thought you just did the same mistake than the p1 dump, actually you did but that's not the biggest issue here, the dump with crc `0xa8603979` actually comes from another hbmame romset named `fatfury1d` : https://github.com/Robbbert/hbmame/blob/master/src/hbmame/drivers/mgd2.cpp#L451
> I can't accept this for 3 reasons :
> 
>     1. we don't mix dumps from different romsets (even if it ended up improving the hack, which probably isn't the case here, authors don't like this)
> 
>     2. the header in hbmame says this one is a bad dump made with "Mega Game Doctor 2"
> 
>     3. it obviously doesn't have anything to do with the purpose of this hack (enabling hidden characters)

Oh I see, that one was in there alongside the other files so I wasn't sure if it was specifically important for this one,

barbudreadmon:
> I can clean this one up a little and submit a new PR if needed

Yes, i'm waiting for your cleaned up code

PR #589, opened 2020-11-20
Fatal Fury 1 Boss Hack Cleanup

WandeingLagomorph793:
As per suggestion I removed the unneeded parent files for both 033-m1h.m1 and 033-p1.p1, I also added EGCG/EGGT for the imitation PS1 hack for King of Fighters 97 since going by the little that mentions the existence of the hack, the only name I found mentioned was that one.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -145,7 +145,7 @@ static struct BurnDIPInfo ms5pcbDIPList[] = {
 	{0,		0xFE, 0,	2,	  "Stop mode"},
 	{0x00,	0x01, 0x80,	0x00, "Off"},
 	{0x00,	0x01, 0x80,	0x80, "On"},
-	
+
 	// Region
 	{0,		0xFD, 0,	2,	  "Region"},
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Looks fine to me

@JacKc029735 you don't see any issue either ?

WandeingLagomorph793:
Looks like it got merged, so I'm guessing I did it right then?

I've got more of these but don't want to send all of them at once since that could lead to complications, I've got a few from Samurai Shodown, I might do a PR for I and V, 1 had some trouble last time so I'm hoping it goes smoother this time, I double checked all the CRCs and verified the sources to make sure something doesn't go wrong. and V doesn't have any matching CRCs with it's parent rom so literally it's entire folder is just the modified files.

I've got some hacks for schmeisr and asurbus but I don't know which .cpp files those would be and I'm still looking for on that.

dinkc64:
schmeiser is in dataeast/d_rohga.cpp
asura buster is in pst90s/d_fuukifg3.cpp

hint, if you have grep (or mingw installed, grep comes with it), use "grep -ir schmeiser *" in src/burn/drv to quickly find the file you are looking for.

PR #591, opened 2020-11-21
Samurai Shodown Hacks

WandeingLagomorph793:
Here's another set of character hacks, just for Samurai Shodown, so 1 through 5. Just to knock out that whole series.

I looked at how the last one went so I'm hoping to avoid the mistakes from before, so no white space issues, properly tagged for | BDF_CLONE | BDF_HACK |, Trimmed files when needed, so it's just ""045-p1eh.p1" instead of both "045-p1eh.p1 and "045-p1.p1" in the same section, especially for Samurai Shodown 5 since there's no matching files from it's normal rom in that. I also moved them to behind where Fatal Fury was. so in the same section as the other hacks rather than being near the bottom with the demos. I compared this in a couple code editors so I'm hoping there isn't a ton of space issues.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -19170,6 +19170,211 @@ struct BurnDriver BurnDrvFatfury1bh = {
 	0x1000, 320, 224, 4, 3
 };
 
+// Samurai Shodown / Samurai Spirits (Secret Character Hack)
+/* MVS VERSION */
+
+static struct BurnRomInfo samshoehRomDesc[] = {
+  { "045-p1eh.p1",  0x100000, 0xba5277df, 1 | BRF_ESS | BRF_PRG }, //  0 68K code 		/ TC538200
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Too many errors i explained in the other PRs, closing.

WandeingLagomorph793:
> Too many errors i explained in the other PRs, closing.

Alright.

How about I just send what I have? I already wasted everyone's time with errors last time and I'd rather not repeat that.

WandeingLagomorph793:
@barbudreadmon 
https://mega.nz/file/14oxxSha#5DoJOT9Te7IEWU7vq8Aa_U40iVCWTaCLXym_jrdHxG8
Don't know if this is the best place to send the link but i figured filing a full issue with just that would be inappropriate and disruptive. That's all the hacks i had alongside a text document with their sources. Samurai Shodown 1 has two of them, they're meant to do the same thing but they both come from different sources, so if CRC comparisons are needed then there's both of them just in case.

barbudreadmon:
I already explained everything that need to be explained, i'm sure you can manage if you read everything carefully.

Please don't post illegal links to copyrighted content.

PR #594, opened 2020-11-23
d_nes.cpp - Big Endian fix

crystalct:


barbudreadmon:
how does it perform on ps3 ?

crystalct:
Really unplayable ... 
image

crystalct:
10-20 FPS

crystalct:
An 8 bits so slow? It's so weird....

mobasuite:
Your running NES on a linux computer.

PR #598, opened 2020-11-24
Neo Geo Hacks

WandeingLagomorph793:
only doing three of these, galaxy fight, art of fighting and world heroes perfect, tried looking through the file set for the remaining titles like d_toaplan but given how everything's gone I'm just sticking with this. Just to run down what i did.

I checked with both the Atom code editor, githubs code editor and CodeSandbox, just to make certain that there wasn't any white space issues, everything should be fine there.

I compared checksums to HBMame, I don't have a single working system where I can run HBMame but I tried at least comparing checksums, I went with these three since no Samurai Shodown game had anything similar within that and since that series only ever presents problems I scrapped that. All three had BH under the header so I matched it according to that for good measure.

I removed parents and replaced them accordingly, so P1BH instead of PHB1 and P1, I also still applied the headers for | BDF_CLONE | BDF_HACK |, and I made sure they went before the Diagnostic Cartridge so they were with the rest of the hacks. This is alongside making sure I have the right headers where they should be and everythings named accordingly.

Had a weird error with github so I had to repull from scratch.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -19170,6 +19170,119 @@ struct BurnDriver BurnDrvFatfury1bh = {
 	0x1000, 320, 224, 4, 3
 };
 
+// Art of Fighting / Ryuuko no Ken (Boss Hack)
+/* MVS AND AES VERSION */
+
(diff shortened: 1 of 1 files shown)


SumavisionQ5:
If those boss hacks can be replaced with UNIBIOS's cheat function, I don't think there is a need to add them.

PR #599, opened 2020-11-24
Three Hacks for Neo Geo

WandeingLagomorph793:
only doing three of these, galaxy fight, art of fighting and world heroes perfect, tried looking through the file set for the remaining titles like d_toaplan but given how everything's gone I'm just sticking with this. Just to run down what i did.

I checked with both the Atom code editor, githubs code editor and CodeSandbox, just to make certain that there wasn't any white space issues, everything should be fine there.

I compared checksums to HBMame, I don't have a single working system where I can run HBMame but I tried at least comparing checksums, I went with these three since no Samurai Shodown game had anything similar within that and since that series only ever presents problems I scrapped that. All three had BH under the header so I matched it according to that for good measure.

I removed parents and replaced them accordingly, so P1BH instead of PHB1 and P1, I also still applied the headers for | BDF_CLONE | BDF_HACK |, and I made sure they went before the Diagnostic Cartridge so they were with the rest of the hacks. This is alongside making sure I have the right headers where they should be and everythings named accordingly.

No idea why it wasn't building right, I went back in and looked all over the file and this time there absolutely shouldn't be any errors from APV, I looked through and added the headers that were needed, I could swear I put those in.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -19170,6 +19170,118 @@ struct BurnDriver BurnDrvFatfury1bh = {
 	0x1000, 320, 224, 4, 3
 };
 
+// Art of Fighting / Ryuuko no Ken (Boss Hack)
+/* MVS AND AES VERSION */
+
+static struct BurnRomInfo aofbhRomDesc[] = {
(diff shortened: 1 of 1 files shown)


WandeingLagomorph793:
Alright I don't know why it's not building correctly anymore, it did before and I checked extra careful to make sure it was working the way it should and make sure it was formatted right. But I won't waste everyones time once more over my own errors so I won't open another one of these.

PR #613, opened 2020-12-02
SDL2: add ALT-ENTER to switch mode fullscreen/window

BeWorld2018:


BeWorld2018:
Fix here : https://github.com/finalburnneo/FBNeo/pull/618

PR #626, opened 2020-12-12
Updated SF 2 Mix

AyanamiRei1:
Changed the Classic Mode dipswitch into Complete as the creator of the hack made the fake 3D and bonus stages appearing as actual stages optional.

The creator of the hack didn't choose the smx naming scheme so I changed that back to the way he intended it to be also updated the crc's as well.


Igor-Arabe:
I think this scheme (smx*) is valid only for emus with native support (fbneo...).
The creator of hack keeps original scheme because is more easy to run replacing original sf2ce (just renaming 'sf2mix' to 'sf2ce') on emulators without support.

dinkc64:
@AyanamiRei1 
Please keep the smx rom prefix :)

best regards,
- dink

PR #643, opened 2020-12-19
1:1 pixel aspect ratio for vector games

barbudreadmon:
all those vector games don't have smooth lines at the moment, this is especially visible if you use the hires mode and look at some scrolling (watch letters in starwars intro, or landscape in bzone), setting pixel aspect ratio to 1:1 (display aspect ratio to width:height) everywhere seems to fix this issue.

btw, i did this PR without merging it because i'm not sure about this solution, my other theory would be that current 4:3 display aspect ratios are actually fine, but the line drawing algorythm need to somehow account for it.


dinkc64:
That's a good idea, let's try it - we can always fake aspect ratio by changing the resolution & pixel aspect ratio, if things look weird

PR #675, opened 2021-01-09
Fix the bug that the defense value of kov2pfwll clothes is wrong

eziochiu:
before fix
5A6B9F18-7345-4EEE-AADB-7C2A56518A8B

after fix：
BB5BB584-F68D-4E52-AFA1-4D9AC3D9DEDA


barbudreadmon:
thanks for your contribution !

PR #690, opened 2021-01-21
cps3: sfiii - sync dip switch to use sfiiiwide dip info

whatdoineed2do:
sync 990680 and 990512 roms' dip info handling as per 970930 to enable normal/wide display mode


barbudreadmon:
Not enabling this widescreen mode for sfiii3 is intentional, since it's way too buggy

PR #694, opened 2021-01-25
fix olds input rank crash except world region [eziochiu]

barbudreadmon:


barbudreadmon:
@eziochiu could you check that i didn't mess this PR while cleaning it ?

eziochiu:
I've checked, It's correct.

eziochiu:
@barbudreadmon I've checked, It's correct.

PR #698, opened 2021-02-02
New Megadrive hack

Charles32k:
Monster world IV french


barbudreadmon:
i'm a bit confused about the nes changes in that PR.
it appears some changes were made to the robot-generated part, that's no good.

PR #712, opened 2021-02-16
fix olds error and drgw3 & drgw3's clone works perfect now!

eziochiu:


barbudreadmon:
Thanks again for your contribution :)

PR #716, opened 2021-02-18
move kaillera chat key

barbudreadmon:
A user with hand disability asked if we could move that key to make things nicer for keyboard users. Is that change fine for everyone ?

### DIFF
--- a/src/burner/win32/run.cpp
+++ b/src/burner/win32/run.cpp
@@ -640,7 +640,7 @@ int RunMessageLoop()
 								}
 								break;
 							}
-							case 'T': {
+							case VK_OEM_2: {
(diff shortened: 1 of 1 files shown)


tmaul:
Sounds reasonable to me, although I'm not a netplay user.

tmaul (approved):

dinkc64:
VK_OEM2 is the / key, right?  I like that idea.  But:
we have to update any references to it.  Is it in the manual, if so, change that also.

barbudreadmon:
Yes, that's the /? key.

fba.chm is totally blank for me ?

mobasuite:
why not update fba.chm to fbn.chm, thx

PR #726, opened 2021-02-24
spectrum: new games.

Svipur:


barbudreadmon:
i think you messed up formating in this one

Svipur:
> i think you messed up formating in this one

Yeah, soz. Cleaned it up.

Issue #743, opened 2021-03-31
[NeoGeo] Inserting coins while playing games makes the inputs drop for a while


ez2torta:
I was trying KOF98 on FinalBurnNeo as most of the tournaments are in Fightcade2.. and there's this problem that FBNeo has but Mame does not have.

Whenever a player presses Coin, the inputs stop responding for a brief second.. i recorded the same in latest Mame but does behave quite well.

How to replicate:
- Load KOF98
- Choose any character in advanced mode
- Start running (tap forward then hold forward) 
- While running press coin button
- Kyo should keep running

FBNeo Behaviour (latest one downloaded from github releases.. as of now 1.0.0)
https://youtu.be/Up_jtUBbrGE

Mame Behaviour (latest one from mamedev.org.. as of now 0229b)
https://youtu.be/HBIThR4JDRI

I would like this to be fixed as it can affect players while doing combos (the receiving player just need to mash coins to make 1P drop combos)

Thanks!

barbudreadmon:
Are you using a keyboard ? Your issue sounds like keyboard's "key ghosting" to me (it's an issue where you can't press several keys at the same time), as a matter of fact i can't reproduce this with a controller. See https://drakeirving.github.io/MultiKeyDisplay/ to test the limits of your keyboard.

If you are using a keyboard, you might have to remap things in a way that won't trigger key ghosting, i don't have a list but i know keys like QWEA are usually fine.

ez2torta:
Hmm strange.. what controller are you using? I do not have key ghosting as i was using the same keyboard in both Mame and FBNeo..

also i tested SFIII3rd and does not have the coin problem :) so it must be just the NeoGeo Driver

dinkc64:
I know why this is happening, inserting a coin causes the input bank to change, but then it tries to read the joy1 when the bank is changed, and reads nothing, then when the game is happy that a coin has been inserted, it sets the inputbank back to joy1.
The trick is coming up with a fix that doesn't break anything else.  let's see...

dinkc64:
Oh dear, this is going to be a real bitch to fix.  Can you (or anyone) help me find another neogeo (aes, mvs) game where this problem also happens?
Sometimes it's helpful to have a second or third problem example, it might give me another lead, because right now I'm out of ideas.

ez2torta:
Hmm i tested kof2002 and it was the same thing :(

[...]

PR #745, opened 2021-04-02
Feature/neogeo coins problem

ez2torta:
Added a condition for the Coin Lockout Problem.. it seems that it switched Input Banks for whatever reason..

The only condition that the input bank was changed/shifted was the SNK_4_JOYSTICKS case.

Also SOCD would be appreciated for keyboard players. It's not a must but with a few lines of code i got it running on NeoGeo :)

Needs testing in more games.. i can't test all of them :/

### DIFF
--- a/src/burn/drv/neogeo/neo_run.cpp
+++ b/src/burn/drv/neogeo/neo_run.cpp
@@ -2097,7 +2097,12 @@ static void WriteIO1(INT32 nOffset, UINT8 byteValue)
 		case 0xE1:
 //			bprintf(PRINT_NORMAL, _T("  - Chute 2 coin lockout -> Low / Input bank 1 selected (0x%02X).\n"), byteValue);
 // 			This was previously + 8, fixes MVS Fighting Games coin issue!
-			NeoInputBank = NeoInput + 0;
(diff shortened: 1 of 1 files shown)


src/burn/drv/neogeo/neo_run.cpp
```diff
@@ -4513,8 +4519,14 @@ INT32 NeoRender()
 
 inline static void NeoClearOpposites(UINT8* nJoystickInputs)
 {
+	// 0x03 to Binary = 00000011 (Up and Down enabled)
+	// 0x0C to Binary = 00001100 (Left and Right enabled)
+	// SOCD Option in Inputs config would be appreciated.
+	// Standard SOCD Config is Up + Down = Up, Left+Right = Neutral
+	bool SOCD = false;
```
FelipeLema, line 4526: you can also use `static constexpr bool SOCD = false` to help the compiler recognize the `if` branch below as "always execute"

FelipeLema:
fixes #743

dinkc64:
I felt compelled to commit my wip from 2 days ago.  but gave you credit in the commitlog since we arrived at the same conclusion.  Hopefully the other input methods are fine (I didn't test much)

PR #757, opened 2021-04-15
d_cop01.cpp: Fix Mighty Guy some sound issues

sasuke-arcade:
Fixed various sound issues in Might Guy so that it sounds almost exactly like PCB.

- BGM tempo was unstable.
  - The frequency setting of the DAC and the clock of the timer are now linked.
  - When the sound driver changed the DAC's clock, the loop count ($C010) that calls the timer in anticipation of this action was fluctuating in the range of 2 to 4 times.
- The playback frequency of the DAC is incorrect.
  - I changed the frequency so that the sound plays as per the PCB.
- Implemented a protect command 0x11 to stop the DAC.
  - This is needed to stop the DAC when the psycho gun is stoppedv
- Bug in which some DAC sound effects did not sound correctly.
  - This is not in MAME, but seems to be a problem when it was ported to FinalBurn.
  - The DAC data reference address mask was wrong.

### DIFF
--- a/src/burn/drv/pre90s/d_cop01.cpp
+++ b/src/burn/drv/pre90s/d_cop01.cpp
@@ -45,6 +45,7 @@ static UINT16 prot_dac_current_address;
 static UINT16 prot_dac_freq;
 static UINT8  prot_dac_playing;
 static UINT8  prot_const90;
+static UINT8  prot_timer_rate;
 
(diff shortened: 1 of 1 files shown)


dinkc64:
sasuke-arcade beautiful work, buddy!! :)
Thank you!

best regards,
- dink

PR #783, opened 2021-05-08
Update d_pgm.cpp

SumavisionQ5:
Do a little modification to kov2pfwllInit so that old kov2pfwll ROMs (or IPS made with old kov2pfwll ROMs) can run correctly.


SumavisionQ5:
In fact, this kov2pfwllinit can be removed. I have a better way to remove the protection.  We no longer need to patch it every time when the ROM is updated.

SumavisionQ5:
I will submit a new commit soon.

SumavisionQ5:
After digging it for a while, I found my new solution is not suitable for latest kov2pfwll. The 68K Code is alreay expanded to 6 MB.  So please forget this commit if you are not going to run old kov2pfwll.

PR #807, opened 2021-07-08
Street Fighter III HBMame

MavisCelus19201:
sourced from hbmame down to checksum and filenames, looked to try and fix quotes, parenthesis and commas compared to last time, my system isn't strong enough to compile without crashes so it's hard to test. not sure on parameters so i'm looking back and forth for comparison.

### DIFF
--- a/src/burn/drv/cps3/d_cps3.cpp
+++ b/src/burn/drv/cps3/d_cps3.cpp
@@ -183,7 +183,7 @@ static struct BurnDIPInfo sfiiiwideDIPList[] = {
 	{0x1B,	0x01, 0x0F,	0x07, "Oceania"},
 	{0x1B,	0x01, 0x0F,	0x08, "Asia"},
 	{0x1B,	0x01, 0x0F,	0x00, "XXXXXX"},
-	
+
 	{0,		0xFD, 0,	2,	  "Screen mode"},
 	{0x1B,	0x01, 0x80,	0x80, "Wide"},
 	{0x1B,	0x01, 0x80,	0x00, "Normal"},
@@ -330,7 +330,7 @@ static struct BurnRomInfo sfiiiRomDesc[] = {
 
 STD_ROM_PICK(sfiii)
 STD_ROM_FN(sfiii)
-	
+
 // -----------------------------------------------
 // Street Fighter III: New Generation (USA 970204)
 // -----------------------------------------------
@@ -367,7 +367,7 @@ STD_ROM_FN(sfiiia)
 static struct BurnRomInfo sfiiijRomDesc[] = {
 
 	{ "sfiii_japan.29f400.u2",				0x080000, 0x74205250, BRF_ESS | BRF_BIOS },	// SH-2 Bios
-	
+
 #if !defined ROM_VERIFY
 	SFIII_970204_FLASH
 #endif
@@ -397,7 +397,7 @@ STD_ROM_FN(sfiiih)
 static struct BurnRomInfo sfiiinRomDesc[] = {
 
 	{ "sfiii_asia_nocd.29f400.u2",			0x080000, 0xca2b715f, BRF_ESS | BRF_BIOS },	// SH-2 Bios
-	
+
 	SFIII_970204_FLASH
 };
 
@@ -410,7 +410,7 @@ STD_ROM_FN(sfiiin)
 static struct BurnRomInfo sfiiinaRomDesc[] = {
 
 	{ "sfiii_asia_nocd.29f400.u2",			0x080000, 0x73e32463, BRF_ESS | BRF_BIOS },	// SH-2 Bios
-	
+
 	SFIII_970204_FLASH
 };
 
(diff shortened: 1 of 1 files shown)


barbudreadmon:
>my system isn't strong enough to compile without crashes so it's hard to test

I can't imagine how it could be possible since even a old raspberry pi model can build this. This PR won't build because part of the issues i mentioned last time are still there.
We won't accept those PRs as long as you don't test it locally and make sure everything works properly (meaning it builds and the hidden characters are actually available).

MavisCelus19201:
> > my system isn't strong enough to compile without crashes so it's hard to test
> 
> I can't imagine how it could be possible since even a old raspberry pi model can build this. This PR won't build because part of the issues i mentioned last time are still there.
> We won't accept those PRs as long as you don't test it locally and make sure everything works properly (meaning it builds and the hidden characters are actually available).

my only system is a weak 32bit windows system and i've had trouble getting githubs tools working with that, not sure if it's too weak or just the 32bit archetecture, though in a few months i'm hoping to upgrade, I'll try getting it working in the meantime though it's been a few years since id messed with this, so apologies if i'm rusty.

dinkc64:
Hi MavisCelus19201,
Thanks for the PR, I found 2 problems that need to be fixed before we can merge it:
1: line 2174, "BDF_HACK, BDF_HISCORE_SUPPORTED", remove the ", BDF_HISCORE_SUPPORTED"
2: line 2171, the first entry in the BurnDriver struct is the romset name, that needs to be changed to "sfiii2bh".  The second entry needs to be "sfiii2", which is the parent romset name (since it is a clone, it needs a parent link)

when you're done, the struct should look like this:
struct BurnDriver BurnDrvSfiii2bh = {
	"sfiii2bh", "sfiii2", NULL, NULL, "1997",
	"Street Fighter III 2nd Impact: Giant Attack | Boss Hack\0", NULL, "hack", "CPS-3",
	NULL, NULL, NULL, NULL,
	BDF_GAME_WORKING | BDF_16BIT_ONLY | BDF_CLONE | BDF_HACK, 2, HARDWARE_CAPCOM_CPS3, GBF_VSFIGHT, FBF_SF,
	NULL, sfiii2bhRomInfo, sfiii2bhRomName, NULL, NULL, NULL, NULL, cps3InputInfo, sfiiiwideusaDIPInfo,
	sfiii2Init, cps3Exit, cps3Frame, DrvCps3Draw, cps3Scan, &cps3_palette_change, 0x40000,
	384, 224, 4, 3
};

best regards,
- dink

dinkc64:
p.s. I have a tiny mingw package that can compile fbneo, it even works on my old p4/winxp machine, if you're interested I can find a link.

MavisCelus19201:
> p.s. I have a tiny mingw package that can compile fbneo, it even works on my old p4/winxp machine, if you're interested I can find a link. All you have to do is unzip it, unzip the latest source and run a batch file.

That'd be nice, testing on a weak 32bit system hasn't been simple, it's been a few years since i've messed with this so i'm a little rusty.

I'll try fixing the structure and see if that works, if not then I can try sending checksums if need be, I can confirm the hacks are working on actual HBMame itself so I'm just having trouble putting them into the right structures.

dinkc64:
[mmingw_2.zip](https://github.com/finalburnneo/FBNeo/files/6789045/mmingw_2.zip)

Unzip it, unpack the rar so it makes c:\mingw
grab the latest fbneo code from the green button on our github page "download zip", unzip it to c:\mingw\fbneo  (or somewhere in there)

now from a cmd window:
run \mingw\bin\mingw_start.bat

in the same cmd window:
go to the fbneo code directory where the makefile resides, and read the message from mingw_start.bat

if you want the compile to go a bit faster, edit makefile.mingw and find the line "DEPEND = 1"
change it to "#DEPEND = 1" & save.  do this before the step above!

best regards,
- dink

barbudreadmon:
>I've also tested the hacks in HBMame itself and they do work, I can send screenshots or video if you want any specific proof but I can confirm on that end they work

As i already said, they need a custom loading mecanism which FBNeo doesn't have, i'll link again that custom loading mecanism from HBMAME : https://github.com/Robbbert/hbmame/blob/master/docs/release/src/hbmame/drivers/cps3.cpp#L35-L192

dinkc64:
barbudreadmon, imo he did the hard part is adding the rom struct, I hate doing that.  so that's positive.
If he can't figure out the unscramble part, i'm sure it can be done easily on my side.

JacKc029735:
ips patches using actual MAME and FB Neo sets already exists for those boss hacks and prefectly works. So it is probably useless to add them.

MavisCelus19201:
> ips patches using actual MAME and FB Neo sets already exists for those boss hacks and perfectly works. So it is probably useless to add them.

you mostly see janky prepatched roms or HBmame more than ips for these, they're also usually strict enough that patching doesn't work without adding support, there's a set of battle garegga patches supported in HBMame that didn't work with FBN when i tried them to name a quick example. some of them offer more to them than MAME cheats offer or just work better, the street fighter 3rd strike one from this set allowing Shin Akuma to name an example

MavisCelus19201:
so I got the file ready and fixed, when I tried compiling it manually it didn't throw any errors when It got to the cps3.cpp, is that a good indicator or should I wait till it finishes so I can try the hack out from within FBN? I'm thinking of doing the latter but It's taken about 3 hours to compile so far so I don't know if looking out for errors on the cps section will be faster.

dinkc64:
Just let it keep going until it finishes :)
After it's done compiling everything, it should only take a minute or so when you make a change to a single file.

PR #821, opened 2021-07-19
crc fix :  Street Fighter II' - Hyper Champion Edition (Hack, v2)

mobasuite:
makes it playable


barbudreadmon:
No issue using the previous one with crc `0xe5c620a9` here, actually that's the crc you'll get if you follow instructions from http://www.romhacking.net/hacks/6060/ so i think this PR must be closed, except if someone can confirm there is actually an issue

PR #824, opened 2021-07-22
Lua support

JordiRos:
Ported from legacy fba-rr (original authors: mauzus, andres.delikat)

Some missing work to do
- added support for vs2015 only, adding support for vs2019 and mingw should be as easy as adding the new files (I don't have access to any of those compilers)
- haven't tested all blitters (pc limitations)
- some luas have dll dependencies (those with image loading mostly)
- haven't added command line options to automatically open luas
- hotkeys are mapped to ALT + 1 to ALT + 5, ideally those could be mapped like any other input|


tmaul:
Cool, thanks!

I will take a look at getting this building with VS2019, and maybe mingw if I have time and no one else beats me to it...

dinkc64:
Very nice, anyone have a link to fba-rr source so I can see what must be done here, and some sample/test lua's?

barbudreadmon:
Probably came from https://github.com/TASVideos/fbarr

JordiRos:
This is the original source
https://code.google.com/archive/p/fbarr/source/default/commits
Usage instructions
https://code.google.com/archive/p/fbarr/wikis/LuaScriptingFunctions.wiki

For it to build correctly, project/make fixes are needed for vs2019 and mingw. Rest of fixes are optional. You may or may not want to disable LUA for online play, or in libretro builds.

To use lua, launch a game, go to File -> Lua Scripting -> New Window, click Browse, select a LUA file and click Run.
Examples
https://www.youtube.com/watch?v=IBb7crQSgJE
https://www.youtube.com/watch?v=OU-D3SEBrgw
Any of these scripts might be compatible
https://github.com/Jesuszilla/mame-rr-scripts

You will need DLL dependencies for some of the scripts, unzip next to fbneo.exe.
https://github.com/JordiRos/fightcade-fbneo/blob/master/luadeps.zip

dinkc64:
Without merging it to the main project first, how can I download this pr to my pc without downloading every single file individually?

grant2258:
I found the easiest way just to test is this.
```
git checkout master
git pull
git checkout -b testing
```
then check your pull request number and add .diff to it so in this case 
```
wget https://patch-diff.githubusercontent.com/raw/finalburnneo/FBNeo/pull/824.diff
git apply 824.diff
```
you can also add a remote to do it but its a little more involved

dinkc64:
thanks grant2258!

dinkc64:
I also tried copying the lib and dll from the fba-rr readme over, that also didn't help.  Although we're linking lua directly into the exe, there should be no need for dll.

grant2258:
can you post the win32 exe and the makefile. I dont have a windows dev env installed but can if i need to test it(assuming msys2 can compile it with no issues.)

grant2258:
ok managed to get it to cross compile on linux 

ant@grant-All-Series:~/Source/FBNeo$ wine ./fbneod.exe --
Z:\home\grant\Source\FBNeo\fbneod.exe: no input files given
usage: Z:\home\grant\Source\FBNeo\fbneod.exe [options] [filenames].
Available options are:
  -        process stdin
  -l       list
  -o name  output to file 'name' (default is "luac.out")
  -p       parse only
  -s       strip debug information
  -v       show version information
  --       stop handling options

dinkc64:
grant2258, that did the trick, thanks again!

grant2258:
anytime my good man!

dinkc64:
this is getting merged, makefile fixes will be getting checked in soon

dinkc64:
so, I loaded up a game
opened the lua console, browsed for "input-display.lua" from https://github.com/Jesuszilla/mame-rr-scripts/blob/master/input-display.lua
press "run" and fbneo window disappears.  It crashed so fast my just-in-time debugger didn't catch it :paintbrush: 

ideas?

grant2258:
Well I opened the first spectrum 007 game in the list loaded the script and got this 

Warning: unable to open 'input-modules.lua'
Generic input display: 1-player, 0-button
* Press Lua hotkey 5 to toggle input display.

grant2258:
never mind downloaded the missing script and get this
Generic input display: 1-player, 0-button
* Press Lua hotkey 5 to toggle input display.

not sure about anything lua to be honest not crashing here though in wine

dinkc64:
@tmaul can you give this a try on windows? :)

tmaul:
@dinkc64 For some reason visual studio 2019 has decided to mess up the xaudio2 headers so will no longer compile so as soon as I've figured out how to fix that then I will take a look. I currently have the same problem as the build bots have. I don't get it... Yet

dinkc64:
@tmaul, oh, that sucks!

so, i logged stderr and get this when fbneo dies (i thought it was crashing, but lua is exit()ing)

"PANIC: unprotected error in call to Lua API (bit library self-test failed (use D3DCREATE_FPU_PRESERVE with DirectX))"

dinkc64:
oh, i think its telling me to use that flag with d3d creation? hmm

tmaul:
Yeah it could be. Maybe it only works with certain blotters?

grant2258:
@tmaul its might be this change for cross compiling to work https://github.com/finalburnneo/FBNeo/pull/827/commits/d7069dec8e0a1ebcf3fc78a382abd28c903fbbeb

barbudreadmon:
yes, the cross-compiling stuff is most likely responsible for this, i kinda remember a PR like this was done in the past and it was reverted for that reason.

grant2258:
in this case im not sure its the cause. its just a change change case fir the includes. windows isint case sensative so i really cant see that change causing it starnger things have happened though

grant2258:
I cross compiled the 64 bit version and tested on windows 10 is working for me. So its probably them flags that are need for win32 like dink said. Included the bin if anyone wants to check it out.

[fbneo64d.zip](https://github.com/finalburnneo/FBNeo/files/6874240/fbneo64d.zip)

dinkc64:
Does someone want to work on these bugs?
I got lua to work by just returning true for the bitop test in luaengine.cpp.  Why it's failing:
it's setting a double to 1437217655L, casting it to a uint32 and checking if it's the same value.  I made a little testprog and compiled it with the same flags as fbn uses, and it never gives a bad result for the uint32 casted double.  what's going on here?  The function in src/burner/luaengine.cpp is luabitop_validate().
my testprog:
#include <stdio.h>
#include <stdint.h>

int main() {
        double z = 1437217655L;

        printf("z = %lf  uz = %d\n", z, (uint32_t)z);
        if ((uint32_t)z != 1437217655L) printf("uhoh.\n");
        return 0;
}

First bug:
game needs to be paused when running a script, otherwise it throws up some errors

second bug, possibly the worst:
it only seems to work properly with the dx9 blitter.   Even when I remove lua from the blitters and set it up properly in vid_interface.cpp like it should be in the first place!  ?!  

lastly, I feel like I'm sinking way too much time into this with little outcome.

tmaul:
I can take a look as soon as I can fix my damn build... No idea what ms have done but visual studio 2019 now hates the fbneo source

mobasuite:
@tmaul 

You do know VS2022 preview is available, it's 64bit based.

dinkc64:
@tmaul, that's a bummer!  can you please give me a log from the compiler and I'll try to figure out what the problem could be?

tmaul:
its in the xaudio2.h header, line 20

"This version of XAudio2 is available only in Windows 8 or later. Use the XAudio2 headers and libraries from the DirectX SDK with applications that target Windows 7 and earlier versions."

grant2258:
@dinkc64 

look at this little bit of magic

https://github.com/finalburnneo/FBNeo/blob/635e24b2447bbea0171808a15a12058276159a0f/src/burner/luaengine.cpp#L3434-L3471

change this to
https://github.com/finalburnneo/FBNeo/blob/635e24b2447bbea0171808a15a12058276159a0f/src/dep/libs/lua/luaconf.h#L504

any of these
#elif defined(LUA_NUMBER_INT) || defined(LUA_NUMBER_LONG) || \
      defined(LUA_NUMBER_LONGLONG) || defined(LUA_NUMBER_LONG_LONG) || \
      defined(LUA_NUMBER_LLONG)

grant2258:
> its in the xaudio2.h header, line 20
> 
> "This version of XAudio2 is available only in Windows 8 or later. Use the XAudio2 headers and libraries from the DirectX SDK with applications that target Windows 7 and earlier versions."

just edit the header and remove the check :)

dinkc64:
@grant2258, thanks for the help, but the problem is/was much deeper - directx/d3d is messing with the fpu precision.  Luckily after some research today I came up with a nice solution.
removing the header check from xaudio2.h is a bad idea, reason to follow..


@tmaul can you try to backlevel xaudio2.h from vs2017 or something? :)  From what I heard, the new "xaudio2" is very different than the old one, and is only compatible with newer os's.  by supporting the older version, we support everyone old/new, and that's good :)

dinkc64:
some better news:
after a crazy day with this stuff, I managed to:
> get lua working on every blitter, including fullscreen
> prevent direct3d/draw from messing with the fpu precision.  at one point I remember this was causing neogeo to round cycles differently on certain blitters due to d3d/dd messing with the fpu precision, and it was causing some non-deterministicity w/recordings, netplay and things.

yay!

curious: does fightcade's fbneo work with lua in fullscreen?  anyone know?

wrennnnnn:
@dinkc64 Just tested, it does with directx9 alt

dinkc64:
thanks wrennnnnn

dinkc64:
lua is now 100% functional in the windows standalone, as far as I can tell :)

grant2258:
Al looks well running under wine as well. It might just be a wine issue but when i select the experimental directx blitter i get a guru meditation

PR #829, opened 2021-08-01
Update inp_sdl2.cpp

petrisi:
Fixed the controller axis checking. We have two checks per axis of the controller. I discovered this when troubleshooting why only left and right (both ends of the first axis) were working on 2-axis Retrolink SNES controller.


dinkc64:
thanks :)

PR #840, opened 2021-08-16
Midway: umk3uk update

Igor-Arabe:
umk3uk updated to version 2021-08-06

2021/08/16 ~ 12h10 (utc -3)

### DIFF
--- a/src/burn/drv/midway/d_wunit.cpp
+++ b/src/burn/drv/midway/d_wunit.cpp
@@ -911,11 +911,11 @@ struct BurnDriver BurnDrvUmk3te = {
 };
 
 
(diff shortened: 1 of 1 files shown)


dinkc64:
@Igor-Arabe, KOHR - author of this hack has kindly asked that we update to his special fbneo version when he sends it, and not to use the same romset from mame/hbmame etc.

Igor-Arabe:
> 
> 
> @Igor-Arabe, KOHR - author of this hack has kindly asked that we update to his special fbneo version when he sends it, and not to use the same romset from mame/hbmame etc.

Ah ok.

PR #844, opened 2021-08-31
lib7z update

barbudreadmon:


barbudreadmon:
@0xe1f macos will need `Sha256Opt.c`, however i've no idea about the proper method for adding stuff to pbxproj files

PR #848, opened 2021-09-07
Build Definition Update - VS 2019

Lincolnhsp:
Build definiton conditional to include Visual C++ 2019 Compiler in the build details.


barbudreadmon:
It doesn't build : 
```
src/dep/scripts/build_details.cpp:59:2: error: #else without #if
   59 | #else
      |  ^~~~
src/dep/scripts/build_details.cpp:61:2: error: #endif without #if
   61 | #endif
      |  ^~~~~
Compiling src/burner/win32/bzip.cpp...

```

PR #858, opened 2021-10-06
Updated main.cpp to fix the broken Mortal Kombat audio

matead92:
The Mortal Kombat games would play audio at the wrong speed and pitch when being loaded from via a command line argument (EG: "umk3 -w"). These two lines of code restart the audio subsystem and fix the audio pitch fix for all Mortal Kombat games.


dinkc64:
thank you :)

PR #867, opened 2021-11-13
adds hotkeys for lua console

BornPosthumous:
This pull request adds hotkeys for the lua console  (alt + key described). This should allow less clicking for lua script developers and better workflow.

Was not sure about updating other VS projects for the scrn.h header file or whether this was the best way to approach this feature.  My CPP knowledge isn't the best and this is a complex code base. Let me know how I should fix things if they aren't correct and I will do my best!

### DIFF
--- a/projectfiles/visualstudio-2019/fba_vs2010.vcxproj
+++ b/projectfiles/visualstudio-2019/fba_vs2010.vcxproj
@@ -33,6 +33,7 @@
     <ClInclude Include="..\..\src\burner\win32\d3dkmt_sync.h" />
     <ClInclude Include="..\..\src\burner\win32\resource.h" />
     <ClInclude Include="..\..\src\burner\win32\resource_string.h" />
+    <ClInclude Include="..\..\src\burner\win32\scrn.h" />
     <ClInclude Include="..\..\src\burn\bitswap.h" />
     <ClInclude Include="..\..\src\burn\burn.h" />
     <ClInclude Include="..\..\src\burn\burnint.h" />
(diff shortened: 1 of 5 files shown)


dinkc64:
Hi @BornPosthumous,
I noticed sometimes if you stop a game and load something else, the lua window will sometimes not appear any more unless I exit fbneo and reload it.   Do you have an idea what might be causing this?

best regards,
- dink

BornPosthumous:
I have been looking for insight on this bug since you mentioned it. I have not fixed it but here is what I have found out:
It looks like the call to open the lua script is made in scrn.cpp (case ID_LUA_OPEN), and the HWnd handle still exists.
However SetForegroundWindow does not bring the window to the front. I tried to make sure it was not simply off screen, but that is not the case either.
The handle is still there, not sure if it is stale.

There is another way to reproduce this bug: Have the lua window open and go to fullscreen. Maybe the window handles get mangled when this happens? I don't know much about the windows API but this seems to suggest to me that destroying the lua window explicitly on game load is not an option for a quick fix (though we could try this anyhow). 

Also tried to make sure this issue wasnt at play:
https://stackoverflow.com/questions/19136365/win32-setforegroundwindow-not-working-all-the-time
Tested that but no dice there either.

Unfortunately checking out older checkouts does not fix it, so its not related to my additions. This would've been easier to track if that were the case :(

dinkc64:
Thanks for looking into it! It might be that when the main window gets recreated on game exit, the hwnd goes stale, though, its only a guess.  I will take a stab at it hopefully soon.

best regards,
- dink

dinkc64:
Hi @BornPosthumous,
I just committed a fix for part of the problem here: https://github.com/finalburnneo/FBNeo/commit/be897811bf3ec0db60018b1adcb931cde7cb5bde
Still don't have a solution to the going fullscreen and back again, but will try to think of something :)

best regards,
- dink

BornPosthumous:
Cool! So basically you just null out the hwnd on close (and some diligent cleanup / UX improvement too). Seems like a good stop gap to me! I checked out and built the branch and the fix seems to work.

PR #919, opened 2022-01-21
Do not overwrite pergame config

UntrustedRoot:
If a per game config exists, do not overwrite.
-joy will still overwrite.


UntrustedRoot:
Forgot to tag this as [SDL2]

Let me know if I should submit this a different way, newish to git.

PR #927, opened 2022-02-07
Add: Windows version & CPU Detection

taoenwen:
1、Windows 11 Support
2、CPU Detection enabled for GCC builds
-------------------
VS
image

GCC
image


dinkc64:
very nice!

PR #930, opened 2022-02-11
nes: fix pal noise frequency.

0kmg:


0kmg:
@dinkc64 I assume this works. Sorry I'm lazy and didn't test it, but it's the same as you did for the DMC table. I made the same change in MAME recently. Table values are from the NesDev wiki, of course, including the minor vbl_length one. Here's an example from MAME that it addresses: https://www.youtube.com/watch?v=SzeLS8jBM_s. Cheers!

PR #933, opened 2022-02-13
nes update: t-chi

taoenwen:


dinkc64:
Hi taoenwen,
Did you run into any unsupported mappers that I could add?

best regards,
- dink

PR #935, opened 2022-02-15
win32: x86 - calculate memory accurately

taoenwen:
image


barbudreadmon:
hmmm will it work on ancient compilers & os ?

dinkc64:
barbudreadmon, it should be fine from xp on up

PR #937, opened 2022-02-17
Add Right-Stick support for MidRes, HBarrel

pjft:
It's my first attempt at this - thanks @barbudreadmon for the guidance with the DIP switches.

I added right-stick control support to Midnight Resistance and Heavy Barrel, two games with dial controls for shooting.

The way I did it is potentially sub-optimal - would definitely welcome feedback and guidance in regards to the actual code - but it works. I know this is an area that @dinkc64 worked a lot on in the past, reading the multiple comments in the code, so I'd definitely heed your guidance and recommendations as well :)

If this ends up being accepted and approved, I'll happily attempt something similar for Ikari Warriors (different driver, so I left it alone for now until I understand the best way to do this with you all).

Thanks and keep up the great work!

EDIT: this hasn't been tested in the standalone version, as I don't have an easy way to do so. This is equivalent to the PR submitted on the libretro fork (https://github.com/libretro/FBNeo/pull/975), but it seems they are pretty much in line with one another it seems. Let me know otherwise. Apologies for the trouble.

### DIFF
--- a/src/burn/drv/dataeast/d_dec0.cpp
+++ b/src/burn/drv/dataeast/d_dec0.cpp
@@ -88,13 +88,14 @@ static INT32 DrvCharPalOffset = 0;
 static INT32 DrvSpritePalOffset = 256;
 
 // Rotation stuff! -dink
-static UINT8  DrvFakeInput[6]       = {0, 0, 0, 0, 0, 0};
+static UINT8  DrvFakeInput[14]       = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}; // 0-5 legacy; 6-9 P1, 10-13 P2
 static UINT8  nRotateHoldInput[2]   = {0, 0};
 static INT32  nRotate[2]            = {0, 0};
 static INT32  nRotateTarget[2]      = {0, 0};
 static INT32  nRotateTry[2]         = {0, 0};
 static UINT32 nRotateTime[2]        = {0, 0};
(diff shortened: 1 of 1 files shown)


dinkc64:
Hi,
Nice work - it looks pretty good!  I'll give it a try soon :)

best regards,
- dink

dinkc64:
It's very nice indeed, the only thing I would do is add a dip option to change the behavior of the second stick:
Second Stick
 -> Moves & Shoots
 -> Moves

.. Well, I did it, https://github.com/finalburnneo/FBNeo/commit/4c6fc13694a20093b8eaddb1c52b52f3615760c9
I was curious to try it out :)

If you add this to other drivers (which would be great), please add the dip option :)

best regards,
- dink

best regards,
- dink

dinkc64:
One last thing: I scanned the autofire variables so states/runahead/input recordings stay in sync :)

best regards,
- dink

pjft:
Oh, wow - thank you so much for the prompt reply. I'm happy it was useful! I will definitely look into it for other drivers and if I run into any challenges I'll either submit a draft PR or reach out on Discord. Thank you!

I will make sure to (try to) add a Dip option on other drivers, and I will also look into scanning the variable. I think I understand what your code does by reading it, but if I'm missing anything I'll let you know - or you can also let me know as well :)

Have a great day and keep up the great work.

dinkc64:
You're welcome, I'm glad things worked out nicely!

best regards,
- dink

PR #940, opened 2022-02-20
Add right-stick rotary support to other drivers

pjft:
Following the previous PR, I'm adding the same support to all the other drivers that have support for a "rotate button" to adjust rotary controls. These are (not exclusively, I didn't include some bootlegs and variants):

- d_seta: downtown, calibr50
- d_snk68: ikari3w, searcharu
- d_snk: bermudat, gwar, tnk3, ikari, victroad /// has issues
- d_alpha68k2: timesold, Btlfield
- d_jackal: jackalr, topgunbl
- d_dec8: gondo

Some notes of what I did and what I found:
- Most games work exactly as intended, which is good. However, this implementation surfaced some existing issues with the existing "rotate button" in games from the *d_snk* driver that I could not solve after digging into the code. Specifically the following:

> * Bermuda Triangle (bermudat): When pressing the rotate button and Fire 1, the rotation goes crazy for a bit. If you press the rotate button and click Fire 1 several times you'll see it go into a constant rotation motion. You can currently replicate it in the stable version - going back to fbalpha - by pressing those buttons. It's not a regression, at least not recent, but messes things up with this control scheme in this game. As I said in the other PR, I don't have a proper debug setup - running things on the Pi - but I suspect it might be related to RotateDoTick going horribly wrong, getDistance is probably breaking for some reason. But you might have much better context on the rotation code :)
> * TNK3: Same behavior, but only for Player 2, but in this case you can replicate it by just pressing the rotate button and a direction. Since this is a "non-simultaneous two-player" game, it was hard to find it out.
> * Guerrilla War (Version 1) / gwara: doesn't rotate with the rotate button, so it doesn't rotate here. Couldn't quite figure out why.

- I tried to make variable scanning consistent across all drivers for rotation variables. Not all drivers scanned all variables, so please confirm whether that is the right thing to do. With this change all of these drivers are scanning nRotate, nRotateTarget, nRotateTry, nAutoFireCounter and nRotateTime.
- I added the DIP option to all, per your recommendation.
- I tested the following games in single and 2 player modes. Here are my notes.
```
// d_seta - OK
Calibre 50 - OK, tested 2 players. Quite fun!
Downtown - OK.
// d_snk - Has issues
Bermuda T: fix autofire - it's always rotating. Not regression, but messes things up. // must be related to RotateDoTick going horribly wrong
TNK3: P2 always rotates - not regression.
Ikari Warriors: Works.
Victory Road: works. 
GWar - Done.
GWara - doesn't rotate.
// d_snk68
Search and Rescue: OK
Ikari 3: OK
//
Time Soldiers:ok
// d_dec8
Gondo - OK
// d_jackal
jackalr: OK 
topgunbl: OK
```
- Last but not least, when developing and testing this, one thing became more obvious. If your analog sticks don't have a good deadzone configured and RetroArch starts with the joystick in a non-centered position, it will somehow detect that you're pressing that joystick all the time. Happens more often when your analog stick is tilting slightly to the left, which makes it assume I'm pushing it to the right. It might be my controllers - an original PS3 SixAxis controller - but since this happened in the past in other emulators, I'm just highlighting it here for the sake of transparency.
 
Hope this works and that it is helpful. It was loads of fun, educational (for me), and I found out a couple of games that turned out to be quite fun to play this way, namely Calibre 50!

As always, happy to hear your feedback.

I suppose, after this, I am missing Forgotten Worlds but that is a completely different beast. I don't think it even has support for a "rotate" button, so that's a bit outside of my current league but happy to take pointers :D

### DIFF
--- a/src/burn/drv/dataeast/d_dec0.cpp
+++ b/src/burn/drv/dataeast/d_dec0.cpp
@@ -2501,7 +2501,7 @@ static void SuperJoy2Rotate() {
 	/// let's try not to over-optimize
 	UINT8 FakeDrvInputPort0[4] = {0, 0, 0, 0};
 	UINT8 FakeDrvInputPort1[4] = {0, 0, 0, 0};
-	UINT8 NeedsAutoFire[2] = {0, 0};
+	UINT8 NeedsSecondStick[2] = {0, 0};
 
 	// prepare for right-stick rotation
 	// this is not especially readable though
@@ -2509,12 +2509,14 @@ static void SuperJoy2Rotate() {
 		for (INT32 n = 0; n < 4; n++) {
 			UINT8* RotationInput = (!i) ? &FakeDrvInputPort0[0] : &FakeDrvInputPort1[0];
 			RotationInput[n] = DrvFakeInput[6 + i*4 + n];
-			NeedsAutoFire[i] |= RotationInput[n];
+			NeedsSecondStick[i] |= RotationInput[n];
 		}
 	}
 
 	for (INT32 i = 0; i < 2; i++) { // p1 = 0, p2 = 1
-		if (NeedsAutoFire[i]) { // or using Second Stick
+		if (!NeedsSecondStick[i])
+			nAutoFireCounter[i] = 0;
+		if (NeedsSecondStick[i]) { // or using Second Stick
 			UINT8 rot = Joy2Rotate(((!i) ? &FakeDrvInputPort0[0] : &FakeDrvInputPort1[0]));
 			if (rot != 0xff) {
 				nRotateTarget[i] = rot * rotate_gunpos_multiplier;
@@ -5841,6 +5843,7 @@ static INT32 DrvScan(INT32 nAction, INT32 *pnMin)
 		SCAN_VAR(nRotateTry);
 		SCAN_VAR(nRotateHoldInput);
 		SCAN_VAR(nAutoFireCounter);
+		SCAN_VAR(nRotateTime);
 
 		SCAN_VAR(nExtraCycles);
 	}
--- a/src/burn/drv/dataeast/d_dec8.cpp
+++ b/src/burn/drv/dataeast/d_dec8.cpp
@@ -60,17 +60,18 @@ static UINT8 DrvJoy3[8];
 static UINT8 DrvJoy4[8];
 static UINT8 DrvJoy5[8];
 static UINT8 DrvInputs[5];
-static UINT8 DrvDips[2];
+static UINT8 DrvDips[3];
 static UINT8 DrvReset;
 
 // Rotation stuff! -dink
-static UINT8  DrvFakeInput[6]       = {0, 0, 0, 0, 0, 0};
+static UINT8  DrvFakeInput[14]      = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}; // 0-5 legacy; 6-9 P1, 10-13 P2
 static UINT8  nRotateHoldInput[2]   = {0, 0};
 static INT32  nRotate[2]            = {0, 0};
 static INT32  nRotateTarget[2]      = {0, 0};
 static INT32  nRotateTry[2]         = {0, 0};
 static UINT32 nRotateTime[2]        = {0, 0};
 static UINT8  game_rotates = 0;
+static UINT8  nAutoFireCounter[2] 	= {0, 0};
 
 static struct BurnInputInfo GhostbInputList[] = {
 	{"P1 Coin",			BIT_DIGITAL,	DrvJoy5 + 5,	"p1 coin"	},
@@ -485,6 +486,10 @@ static struct BurnInputInfo GondoInputList[] = {
 	{"P1 Fire 3 (Rotate)", BIT_DIGITAL,	DrvFakeInput + 4,  "p1 fire 3" },
 	{"P1 Rotate Left",	BIT_DIGITAL,	DrvFakeInput + 0,  "p1 rotate left" },
 	{"P1 Rotate Right",	BIT_DIGITAL,	DrvFakeInput + 1,  "p1 rotate right" },
+	{"P1 Shoot Up"       	, BIT_DIGITAL  , DrvFakeInput + 6,  "p1 up 2" }, // 6
+	{"P1 Shoot Down"      	, BIT_DIGITAL  , DrvFakeInput + 7,  "p1 down 2" }, // 7
+	{"P1 Shoot Left"       	, BIT_DIGITAL  , DrvFakeInput + 8,  "p1 left 2" }, // 8
+	{"P1 Shoot Right"      	, BIT_DIGITAL  , DrvFakeInput + 9,  "p1 right 2" }, // 9
 
 	{"P2 Coin",			BIT_DIGITAL,	DrvJoy5 + 6,	"p2 coin"	},
 	{"P2 Start",		BIT_DIGITAL,	DrvJoy4 + 1,	"p2 start"	},
@@ -497,19 +502,26 @@ static struct BurnInputInfo GondoInputList[] = {
 	{"P2 Fire 3 (Rotate)", BIT_DIGITAL,	DrvFakeInput + 5,  "p2 fire 3" },
 	{"P2 Rotate Left",	BIT_DIGITAL,	DrvFakeInput + 2,  "p2 rotate left" },
 	{"P2 Rotate Right",	BIT_DIGITAL,	DrvFakeInput + 3,  "p2 rotate right" },
+	{"P2 Shoot Up"       	, BIT_DIGITAL  , DrvFakeInput + 10, "p2 up 2" },
+	{"P2 Shoot Down"      	, BIT_DIGITAL  , DrvFakeInput + 11, "p2 down 2" },
+	{"P2 Shoot Left"       	, BIT_DIGITAL  , DrvFakeInput + 12, "p2 left 2" },
+	{"P2 Shoot Right"      	, BIT_DIGITAL  , DrvFakeInput + 13, "p2 right 2" },
 
 	{"Reset",			BIT_DIGITAL,	&DrvReset,		"reset"		},
 	{"Dip A",			BIT_DIPSWITCH,	DrvDips + 0,	"dip"		},
 	{"Dip B",			BIT_DIPSWITCH,	DrvDips + 1,	"dip"		},
+	// Auto-fire on right-stick
+	{"Dip C", 			BIT_DIPSWITCH,	DrvDips + 2, 	"dip"       },
 };
 
 STDINPUTINFO(Gondo)
 
 static struct BurnDIPInfo GondoDIPList[]=
 {
-	DIP_OFFSET(0x17)
+	DIP_OFFSET(0x1f)
 	{0x00, 0xff, 0xff, 0xff, NULL					},
 	{0x01, 0xff, 0xff, 0xef, NULL					},
+	{0x02, 0xff, 0xff, 0x00, NULL                   },
 
 	{0   , 0xfe, 0   ,    4, "Coin A"				},
 	{0x00, 0x01, 0x03, 0x00, "2 Coins 1 Credits"	},
@@ -550,6 +562,11 @@ static struct BurnDIPInfo GondoDIPList[]=
 	{0   , 0xfe, 0   ,    2, "Allow Continue"		},
 	{0x01, 0x01, 0x10, 0x10, "No"					},
 	{0x01, 0x01, 0x10, 0x00, "Yes"					},
+
+	// Dip 3
+	{0   , 0xfe, 0   , 2   , "Second Stick"           },
+	{0x02, 0x01, 0x01, 0x00, "Moves & Shoots"         },
+	{0x02, 0x01, 0x01, 0x01, "Moves"                  },
 };
 
 STDDIPINFO(Gondo)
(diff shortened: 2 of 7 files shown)


dinkc64:
Awesome work! :)
I will try to fix some of the problems you've mentioned, unfortunately though, tnk3 and calibr50 have some problems that I can't solve ATM.  tnk3 being the p2 issue and calibr50 goes bonkers when you get in the aircraft and shoot, as the aircraft turns.  even though the aircraft scene is very short, I'd like to try to come up with a solution someday.

p.s. thanks for upgrading the dips of the games' being changed to the DIP_OFFSET style :)

if you don't mind, I'd like to try to fix the problems before merging this PR, so I don't get too confused by the new code.

best regards,
- dink

dinkc64:
Here's the fix for gwara, pre90s/d_snk.cpp:
find GwaraInit() and before the DrvDoReset() line, put in this line: (see GwarInit() above)
	RotateSetGunPosRAM(&DrvSprRAM[0x3d3], &DrvSprRAM[0x437], 2);

dinkc64:
here's a diff to the original d_snk which fixes: gwara, bermudat & worldwar
[diffy.txt](https://github.com/finalburnneo/FBNeo/files/8106341/diffy.txt)

pjft:
By all means, thank you for the review and for digging into the code :) I just applied the patch that you sent - thanks.

I admit I didn't play calibr50 enough to get so far, but I'll look into it and see what it is and if I come up with anything.

In regards to merging only when those problems are fixed, of course, I have no issues with that. If it ends up being too hard, I am happy to split the PR.

A question, that may help me also debug some of those things: what's the easiest way for me to add classic printf logs to the command line from inside those files there?

dinkc64:
well, tnk3 is next and it's the biggest pain, I've almost got it cracked though!  When it's done we should be good.  

We can save this one for later:
I might give another try at caliber50's aircraft scene though - I had tried several times before without luck (the game doesn't say *in memory* which direction the turret on the plane is pointing)

bprintf(0, _T("x")); or dprintf(_T(""));  or dprintf(L""); will do a debug printf to the console window and logfile on the windows (main) version, I think it also adds to the log on the libretro version.  If you're using the sdl version you could probably just use printf(), but make sure to remove them before adding to the pr.  I'm only 100% sure about the windows version, since that's the one I work on.

best regards,
- dink

pjft:
Sounds good, thanks for the pointers - I am on Linux, Libretro, but will give it a shot. 

Following your comments, I tried out calibr50 earlier and I didn't get to a section inside an aircraft? My playthrough was similar to https://www.youtube.com/watch?v=i1Ti2DCj6jI and https://www.youtube.com/watch?v=hXLjvCkW1wM, so I might have missed a branching path somewhere. If you want to share how to get there I'll try to see what it's all about.

That being said, one thing I noticed - which was interesting and worth mentioning - is that the dial in this game seems to have 16-24 directions, rather than 8 only, making it similar to Forgotten Worlds in terms of "granularity" of the dial. I noticed that because somehow I managed to get "stuck" aiming for a position that wasn't one of the "expected" 8 directions (I probably died while I was turning, and came back in that position). What that meant was that after that misalignment, every time I tried to aim in one of the directions, the aim would never settle there - alternating quickly between the two adjacent positions in a 16-direction chart (so, for simplification, imagine that the directions are the positions of the hours in a clock - 12 directions. I would be trying to aim to the left - 9 o'clock - but it would quickly alternate between 8 and 10, never fully stopping at 9). I'll see if I can dig into that, but I assume it might be potentially related to the "step" of the rotation? I mean, I know I'm talking about this without having so much _seen_ the actual code working step by step, but that's where I'll start looking into it.

I also need to adjust the Calibr50 shooting pattern - by reducing the on-off period cycle for Downtown, I might have made for some issues in Calibr50 with later weapons, specifically the flamethrower. Given that in Calibr50 you can just hold the fire button and don't need to press it multiple times, I'll adjust that and test it out later. Don't merge before we're ok with that. :)

Sorry for the trouble, and thanks for the availability here.

dinkc64:
tnk3 2p fix diff:
[diffy2.txt](https://github.com/finalburnneo/FBNeo/files/8110387/diffy2.txt)

best regards,
- dink

pjft:
Ok, now I ran into the plane section and I get what you're saying :) Indeed, that seems to be a hard one to solve for.

As for the issue I ran into with the direction, it seems to be gone if the step becomes 1 instead of the current 4, but of course, it then makes turning extremely slow. I'll work something out. Updated the PR with the current state of it. I can squash it afterwards if it helps.

pjft:
So, these are a sample of the logs I get [from here](https://github.com/finalburnneo/FBNeo/blob/75b2840ab27d7c5b77d7717cd171bff5568885be/src/burn/drv/pst90s/d_seta.cpp#L5356) (I added the output of getDistance() to it as well, and removed the player index), when we get stuck in the situation of the aim not stopping where we want it to. In this case, I was aiming to the bottom right, and the player was shooting in the two adjacent directions to _exactly bottom right_, alternating between the two. Unsure if the logs will help in any way.

```
Distance 0 Target 6 mempos 7 nRotate 18 try 0.
Distance 0 Target 6 mempos 7 nRotate 14 try 0.
Distance 1 Target 6 mempos 5 nRotate 18 try 0.
Distance 1 Target 6 mempos 5 nRotate 1C try 0.
Distance 0 Target 6 mempos 7 nRotate 18 try 0.
Distance 0 Target 6 mempos 7 nRotate 14 try 0.
Distance 1 Target 6 mempos 5 nRotate 18 try 0.
Distance 1 Target 6 mempos 5 nRotate 1C try 0.
Distance 0 Target 6 mempos 7 nRotate 18 try 0.
Distance 0 Target 6 mempos 7 nRotate 14 try 0.
Distance 1 Target 6 mempos 5 nRotate 18 try 0.
Distance 1 Target 6 mempos 5 nRotate 1C try 0.
```

I didn't capture this specific example on video, but I have a similar video clip when aiming up. 

https://user-images.githubusercontent.com/22025709/155005526-82436f57-3086-444c-9787-22f2bd34387c.mp4

At the end I try to turn to the other diagonals but you'll see that they're all off by (a tiny bit). I don't quite know how to replicate it consistently. My assumption is that it happens when the player dies during turning, and comes back to life aiming to a position that it shouldn't, but I'm no longer as sure of that.

[I'd like to get the step back to 4 as you had it here](https://github.com/finalburnneo/FBNeo/blob/75b2840ab27d7c5b77d7717cd171bff5568885be/src/burn/drv/pst90s/d_seta.cpp#L5229), but if we are in one such position (i.e. if we need to turn less than the step), we'd just turn the remaining delta.

I tried making steps of 2, but I still got stuck in such a situation at some point, so that doesn't seem to be a reasonable workaround.

If you have any thoughts around it, let me know. Thanks for the availability.

dinkc64:
Hi,
thanks for all the info's, I managed to come up with a real nice solution to calibr50's overshoot in RotateLeft/Right(), just need to do a little impl+testing and I'll post back here :)

best regards,
- dink

dinkc64:
Please give this one a try for d_seta.cpp, I added some notes because I couldn't remember how the whole thing worked o.O
If you can get it to go bonkers again, change the "6 : 2"'s in the RotateLeft/Right func's to "4 : 1".

p.s. those skull-throwing monkeys are quite the change-up, don't you think?
[seta_diff.txt](https://github.com/finalburnneo/FBNeo/files/8114029/seta_diff.txt)

pjft:
Wow, thank you for the awesome work and quick response! I'm reading the code, and well done on the changes - it's an elegant way to accomplish what we were aiming for.
Out of curiosity: were you able to pinpoint how many distinct directions the game supports? Reading online I can't find reliable sources - I don't know if it's 16, 24 or another number.
Technically, reading the code I can imagine that it will be possible to get it in such a situation, as if it gets stuck in a state where it's 1, moving it by 2 will keep it alternating between both sides. I also had tried it with 2 and I believe I got it to get stuck in that situation, but I will see how it goes now :) If I get stuck, I'll test 6:1 and 4:1.
I'm about to start my workday now, but I'll test it in more detail later in the day and report back. Thank you!

And yes, those skull-throwing monkeys are an unexpected change stuck in the middle of the game :)

dinkc64:
Thanks, I spent the most time just trying to figure out how this code worked.. Didn't name the variables too sensibly back then.

The game has 16 positions, but each position needs to be clocked 4 times via the input, so the game takes 0 - 3c and moves the guy every 4 increments.

Hopefully the "6 : 2" works out, because it makes the guy move nice 'n fast :)

best regards,
- dink

pjft:
Hi! Apologies for the delay here. :)

So, I tested 6:2 and indeed I ran into the same issue. 6:1 and 4:1 work equally well, it seems - I completed the game twice with them and didn't manage to trigger the issue. I also re-added something that I had lost in the previous diffs that addressed an issue with the flamethrower.

I ended up settling on the 4:1 because, even though it's slower, the angles that the character shoots at during the rotation feel more reliable (every 45 degrees for the most part, where with 6 it felt less reliable to hit enemies in the middle, especially when you're trying to shoot faster around you in multiple directions).

That being said, it is very much a personal preference and I accept either option. I suppose we _could_ add a Dip setting for that, but we might be overengineering it :)

I am happy with where we're at - even the plane section, while not working well, is manageable as it's small. If only "up" aimed the turret to the front of the airplane, it could probably be easier to understand the controls (as it'd always be relative to where the plane is flying to). Right now, however, "up" aims to the plane's left, which just means that I'll always keep shooting in that direction for the most part and it'll destroy the tank.

Is it "easy" to detect if we're in the plane, though? If so, maybe I can add some sort of change to the rotation controls for that occasion (basically, if in plane, force directions to be rotated by 90 degrees, clockwise). 

Regardless, thank you so much for helping get this this far. Let me know how else I can help - do you want me to squash the 3 commits? 

Best regards.

dinkc64:
It's ok, there's definitely no rush :)
I tried with the 4:1 and doing a sweep shoot is better, glad ya noticed that!

Sure you can detect if in the plane, search for "// calibr50 switcheroo", but it's a lot more complicated than just rotating the controls.  It starts out fine, then as the plane rotates the controls get more out of sync with each rotation.  Tonight I'll spend some time trying to figure out the direction of the plane to rotate the controls accordingly.

btw, no need to squash anything, it gives the options to do that when I merge the PR.

best regards,
- dink

dinkc64:
There's an effect that isn't quite right: when entering the holes in the ground, the screen is supposed to go wavey (in typical raster-effect fashion).  I'll try to get that fixed in the next few days.  separate from this pr, of course :)

dinkc64:
woohoo!  after some intense debugging I managed to find the plane's rotation address

dinkc64:
After compensating for the plane's rotation, controlling the plane's turret is excellent! :)
Sometimes the turret jitters if the plane rotates when it's trying to hit the target, I'll have to save fixing that for the morning.

goodnight, best regards,
- dink

pjft:
Oh, wow - I'm looking forward to it :)

One thing I'm thinking, given these new details around the 16-direction dial, is - after this PR - to explore making the shooting be controlled by analog inputs, rather than up/down/left/right digital ones. But I won't do it right now just so we can get closure on this one.

Well done - what a great accomplishment there on the turret :)

dinkc64:
thanks!
Here goes, give this one a try:
[seta_diff_fix_airplane.txt](https://github.com/finalburnneo/FBNeo/files/8125095/seta_diff_fix_airplane.txt)

The analog directionals sounds like a nice feature to have and trivial to impl.  :)

best regards,
- dink

pjft:
I apologize, I have been unable to pick this up today and though I'll try to find some time before the weekend, I won't promise.

You do create high expectations for the analog changes though :D But if we make it happen, I'd like to see if Forgotten Worlds could also follow suit, but from what I observed it did not seem to have a "rotate" button like these, which is the logic I base my changes on (yes, I'm just adding a layer on top of all the hard work you already had done 😅 ).

Just wanted to update this as you've been extremely responsive and I didn't want to come across as not engaging. If I get some time prior to the weekend I'll get back to you, but if not, I'll definitely dig into it in the weekend :)

Thanks again!

dinkc64:
That's not a problem, take your time :)  The changes I made are kinda simple, just a few things to copy over: function adjusted_rotate_gunpos() and replacing *rotate_gunpos[] in 3 places with the function call.

I couldn't do forgotten worlds because at the time *a goofball* wouldn't let anyone touch the cps emulation, he's long gone now, so let's have at it sometime!

best regards,
- dink

dinkc64:
Would it be too much of a hassle if I committed my changes to d_seta?

pjft:
There's absolutely no problem! If you're happy with my changes as well, let's merge and move with it :)

pjft:
Ok, I added them now - though I did not test them yet! :)

pjft:
If you test it and confirm nothing got lost in the translation, I'm fine with it. I doubt I'll be able to test it before end of day GMT. Afterwards I'm happy to give the analog controls a shot in a separate PR - either now or in the weekend :)

dinkc64:
thanks!  I'll merge it sometime today after I clear up a huge mess in d_seta on my side.

best regards,
- dink

pjft:
Thank you! I ended up testing the turret controls earlier, it really makes a difference - well done on figuring it out! :) 

Thanks for all the support in getting this finished - I wouldn't have been able to put it together without your context and knowledge of the code and the games. I hope that this helps more people enjoy these games with more accessible controls. 

I'll try to look into the analog stick during the weekend, if all goes well.

dinkc64:
Thanks, that plane turret has haunted me for ages, it's real nice to finally have that egg cracked. :)

You're welcome, you did a great job on this :)
About a month ago someone on the forum requested this feature, I think they'll be delighted to give it a try.

good luck & best regards,
- dink

dinkc64:
Today was kinda exciting:
I figured out the raster effect - it happens when you go down the hole and when coming back from defeating the monkeys.  It should be checked in soon :)

Did a full playthrough of the game while recording inputs and had great time rambo-ing through the viet-kong army!  The input recording played back entirely in-sync, so that means everything is pretty solid.

Was thinking of detecting the flame thrower weapon and holding the button down instead of toggling it, I think it might help a little in those really insane parts where the soldiers are coming from all sides.  What do you think?

best regards,
- dink

dinkc64:
I just checked in your recommendation - now I can definitely kick more ass in calibr50 with the button being held down instead of pulsed :)
No problem, It's active-low, so removing the bit is fire.  

Regarding your analog-inputs project, if you wanted to figured out everything on your own, please ignore this below :)

...otherwise...

To make things easier, let me tell you about the analog stuff/helper func:
in inputs:
static INT16 Analog[2]; 
..
	A("P1 Stick X",             BIT_ANALOG_REL, &Analog[0],"p1 x-axis"),
	A("P1 Stick Y",             BIT_ANALOG_REL, &Analog[1],"p1 y-axis"),
..
// ProcessAnalog() will convert the analog value to: 0x00 full left, 0x80 center, 0xff full right
// after " / 31": x/y will be 0-8, 4 being center (0123,4,5678)
	INT16 x = ProcessAnalog(Analog[0], 0, INPUT_DEADZONE, 0x00, 0xff) / 31;
	INT16 y = ProcessAnalog(Analog[1], 0, INPUT_DEADZONE, 0x00, 0xff) / 31;

Hopefully I didn't make any mistakes, still waking up here :)

best regards,
- dink

dinkc64:
:) ya welcome!

PR #950, opened 2022-02-27
Add 16-direction analog support to calibr50

pjft:
So, your comment on `Joy2Rotate`

    // ugly code, but the effect is awesome. -dink

is the understatement of the year. Wait until you see `Joy2Rotate16`. :)

I followed your initial pointers and then fumbled my way through it, lots of printfs, a bit of trial and error. Thanks for kicking me off in the right direction :)

Now, this seems to work. It plays "nicer", to be honest. I am not 100% satisfied with parts of the code (it works, but doesn't feel as elegant as _I feel it should be_), and with the gameplay handling of the analogs-to-16-directions feel - I tried different granularity of intervals, but couldn't find one that feels _just right_. The current one seems to work well (it's dividing by 32 rather than 31, I am depending on the dead zone to work as intended, but let me know if that's not correct.

I'll wait for your feedback for now, as I imagine that you may have other pointers.

Also, two moderately unrelated things:
- The changes to the controls, especially the description, were paired with a change to a libretro-only file that's here: https://github.com/pjft/FBNeo/commit/086278bc5ace1de82069042b70bd0648562c9604#diff-6ac1c6d8d8ba7a4551ae7d8b5287044eb322261577f4d23ffe77fca838ecb0c0 . What's the correct process for such a change, as you don't seem to have that file here? @barbudreadmon , maybe you can advise - or, upon merging with the libretro core you can add this change there so that the right-stick gets assigned to the shooting, as an analog input?
- The turret works great, but the moment it changes there's a fraction of time where the gun kind of stays in the previous direction before turning. It makes it as if it's shifting awkwardly for a frame or two, but from reading your code it seems to be doing what it intends to do, in the same order as it should, so I imagine it's probably working as intended, but I thought I'd bring it up in case something occurs to you.

Hope this works, and let me know if there's anything here I should change somehow, or something that I grossly misunderstood from the current code and overengineered things :D 

Have a great Sunday!

### DIFF
--- a/src/burn/drv/pst90s/d_seta.cpp
+++ b/src/burn/drv/pst90s/d_seta.cpp
@@ -1185,10 +1185,8 @@ static struct BurnInputInfo Calibr50InputList[] = {
 	{"P1 Button 1",		BIT_DIGITAL,	DrvJoy1 + 4,	"p1 fire 1"	},
 	{"P1 Button 2",		BIT_DIGITAL,	DrvJoy1 + 5,	"p1 fire 2"	},
 	{"P1 Button 3 (rotate)",		BIT_DIGITAL,	DrvFakeInput + 4,	"p1 fire 3"	},
-	{"P1 Shoot Up"       	, BIT_DIGITAL  , DrvFakeInput + 6,  "p1 up 2" }, // 6
-	{"P1 Shoot Down"      	, BIT_DIGITAL  , DrvFakeInput + 7,  "p1 down 2" }, // 7
-	{"P1 Shoot Left"       	, BIT_DIGITAL  , DrvFakeInput + 8,  "p1 left 2" }, // 8
-	{"P1 Shoot Right"      	, BIT_DIGITAL  , DrvFakeInput + 9,  "p1 right 2" }, // 9
+	A("P1 Aim X", BIT_ANALOG_REL, &DrvAnalogPort0,"p1 x-axis"),
+	A("P1 Aim Y", BIT_ANALOG_REL, &DrvAnalogPort1,"p1 y-axis"),
 
 	{"P2 Coin",		BIT_DIGITAL,	DrvJoy3 + 6,	"p2 coin"	},
 	{"P2 Start",		BIT_DIGITAL,	DrvJoy2 + 7,	"p2 start"	},
@@ -1199,10 +1197,8 @@ static struct BurnInputInfo Calibr50InputList[] = {
 	{"P2 Button 1",		BIT_DIGITAL,	DrvJoy2 + 4,	"p2 fire 1"	},
 	{"P2 Button 2",		BIT_DIGITAL,	DrvJoy2 + 5,	"p2 fire 2"	},
 	{"P2 Button 3 (rotate)",		BIT_DIGITAL,	DrvFakeInput + 5,	"p2 fire 3"	},
-	{"P2 Shoot Up"       	, BIT_DIGITAL  , DrvFakeInput + 10, "p2 up 2" },
-	{"P2 Shoot Down"      	, BIT_DIGITAL  , DrvFakeInput + 11, "p2 down 2" },
-	{"P2 Shoot Left"       	, BIT_DIGITAL  , DrvFakeInput + 12, "p2 left 2" },
-	{"P2 Shoot Right"      	, BIT_DIGITAL  , DrvFakeInput + 13, "p2 right 2" },
+	A("P2 Aim X", BIT_ANALOG_REL, &DrvAnalogPort2,"p2 x-axis"),
+	A("P2 Aim Y", BIT_ANALOG_REL, &DrvAnalogPort3,"p2 y-axis"),
 
 	{"Reset",		BIT_DIGITAL,	&DrvReset,	"reset"		},
 	{"Service",		BIT_DIGITAL,	DrvJoy3 + 5,	"service"	},
@@ -1435,7 +1431,7 @@ STDDIPINFO(Usclssic)
 
 static struct BurnDIPInfo Calibr50DIPList[]=
 {
-	DIP_OFFSET(0x1d)
+	DIP_OFFSET(0x19)
 
 	{0x00, 0xff, 0xff, 0xfe, NULL			},
 	{0x01, 0xff, 0xff, 0xfd, NULL			},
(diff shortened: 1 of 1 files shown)


barbudreadmon:
```
	A("P1 Aim Left/Right", BIT_ANALOG_REL, &DrvAnalogPort0,"p1 Aim Left/Right"),
	A("P1 Aim Up/Down", BIT_ANALOG_REL, &DrvAnalogPort1,"p1 Aim Up/Down"),
```

The strings at the end aren't ok, you need to use strings like `p1 x-axis` and `p1 y-axis`.

Note that this kind of layout (1 dpad + 1 analog) won't automatically be understood as a dualstick layout by the libretro core, so i'll need to add some code for this after merging it.

barbudreadmon:
I'd also recommend using something like "P1 Aim X" and "P1 Aim Y" for description, longer descriptions are not necessarily better :)

dinkc64:
Just had a nice game with this new version of your code, It's really nice! awesome job! :)  I'm at least 10-15% better at roasting the opposition now :)

dinkc64:
I'd just leave everything as-is, except for the input names (first and last fields), as barbudreadmon recommended.
If I can think of ways to improve anything, I'll let ya know!

best regards,
- dink

pjft:
Thanks both - happy that it's increasing your opposition-roasting stats :)

@barbudreadmon Thanks - I made the strings shorter now, Aim X and Aim Y. I had copied the [original suggestions from Wacko's entry in retro_input.cpp](https://github.com/libretro/FBNeo/blob/master/src/burner/libretro/retro_input.cpp#L1504), but I like the shorter ones, which are also used elsewhere there.

Correct, it won't work out of the box. [I made the change to retro_input.cpp that I linked to earlier in my libretro repo](https://github.com/pjft/FBNeo/commit/96240ee9ab5bcbaf818dd720490568e4eb9d5ab5#diff-6ac1c6d8d8ba7a4551ae7d8b5287044eb322261577f4d23ffe77fca838ecb0c0), and it's the following:

```
	// Caliber 50
	if ((parentrom && strcmp(parentrom, "calibr50") == 0) ||
		(drvname && strcmp(drvname, "calibr50") == 0)
	) {
		if (strcmp("Aim X", description) == 0) {
			GameInpAnalog2RetroInpAnalog(pgi, nPlayer, RETRO_DEVICE_ID_ANALOG_X, RETRO_DEVICE_INDEX_ANALOG_RIGHT, description);
		}
		if (strcmp("Aim Y", description) == 0) {
			GameInpAnalog2RetroInpAnalog(pgi, nPlayer, RETRO_DEVICE_ID_ANALOG_Y, RETRO_DEVICE_INDEX_ANALOG_RIGHT, description);
		}
	}
```

I am happy to submit a PR to the libretro core - either before or after the merge - with this change if it's preferable, that was my question. Happy to follow your guidance.

Please let me know if there's anything else needed here!

Have a great weekend.

barbudreadmon:
this is still not right, as i said you need to rename the last string to "p1 x-axis" and "p1 y-axis"

pjft:
@barbudreadmon Apologies, I misunderstood. I thought you had asked for x-axis OR Aim X, and I kind of assumed that the two strings would need to be similar for some reason. Hope it's alright now.

What is each of them used for, by the way? The definition just says szName and szInfo, but it isn't obvious what each is for. the first one seems to show up on the lr-fbneo menu. What's the "x-axis" and "y-axis" used for?

Thanks.

barbudreadmon:
The first one is a description which can be whatever you want, the second one is a key string indicating where the control should be mapped by default.

dinkc64:
After having another play, I really love how this worked out!  Nice work :)

dinkc64:
Howdy,
I did a little re-organization & fixed the rotate button (some ppl don't have dual stick gamepads).  Are you OK with these changes?  https://github.com/finalburnneo/FBNeo/commit/785c6978a898cffd9b4eead7f471fa694f849d0e

I don't really understand the if/else statement under the ProcessAnalog() func, I mean, I can tell it's converting the analog values to signed, but .. that's it :)

bye for now & best regards,
- dink

pjft:
Ah, apologies - did I break the rotate button? I forgot to test it in the end - well done for spotting it. Thanks for that.

I like the re-organization, and without testing it I am mostly fine with those changes. Two notes:
- Here you're running "ProcessAnalog()" for both Downtown and Calibre 50, even though only Calibre50 is the only one that uses the Analogs. Not only is that unnecessary, I _suspect_ that this changes breaks Downtown as it'll set the DrvFakeInput to { 0, 0, 0, 0 } during ProcessAnalog(), rather than keeping the values that it's getting from the joystick. Is there a reason to move ProcessAnalog() from calibr50_input_read? Still, if you tested Downtown and it works, then I don't have a strong opposition to moving it to where it is, though that code that it's running is pointless for any other game. Perhaps at least add an ìf (game_rotate == 1) prior to it.
- The if/else in ProcessAnalog() is really just converting the analog value to "how far it's being pressed in that axis", and assigning the value to the right DrvFakeInput entry. Let me try to explain what I'm going for. So, the current DrvFakeInput array for a single player, the way it's used in Downtown and in all other drivers is (starting at index 6, but keeping it here at 0 for simplicity):

```
DrvFakeInput[ up, down, left, right ]
```

So, since the analog input will return something like:

```
DrvAnalogPort0: left (goes to DrvFakeInput[2]) -> (0123,4,5678) <- right (goes to DrvFakeInput[3]) 
DrvAnalogPort1: up   (goes to DrvFakeInput[0]) -> (0123,4,5678) <- down  (goes to DrvFakeInput[1]) 
```

The logic there aims to return the value towards a direction in a range of (1-4), where 1 is the weakest and 4 is the strongest. Then I can use that to detect differences in those to suggest that the player is aiming at one of the new directions. It will convert, for example:

```
DrvAnalog0 = 0 -> DrvFakeInput[2] = 4
DrvAnalog0 = 1 -> DrvFakeInput[2] = 3
DrvAnalog0 = 2 -> DrvFakeInput[2] = 2
DrvAnalog0 = 3 -> DrvFakeInput[2] = 1
---
DrvAnalog0 = 5 -> DrvFakeInput[3] = 1
DrvAnalog0 = 6 -> DrvFakeInput[3] = 2
DrvAnalog0 = 7 -> DrvFakeInput[3] = 3
DrvAnalog0 = 8 -> DrvFakeInput[3] = 4
---
DrvAnalog1 = 0 -> DrvFakeInput[0] = 4
DrvAnalog1 = 1 -> DrvFakeInput[0] = 3
DrvAnalog1 = 2 -> DrvFakeInput[0] = 2
DrvAnalog1 = 3 -> DrvFakeInput[0] = 1
---
DrvAnalog1 = 5 -> DrvFakeInput[1] = 1
DrvAnalog1 = 6 -> DrvFakeInput[1] = 2
DrvAnalog1 = 7 -> DrvFakeInput[1] = 3
DrvAnalog1 = 8 -> DrvFakeInput[1] = 4
```

The way the rotation is then determined is like this clockwise reference table, starting with up:

- **Up**: x = 0, up = 4
- right < up
- **Up Right**: right = up
- right > up
- **Right**: right = 4, y = 0
- right > down
- **Down Right**: right = down
- right < down
- **Down**: x = 0, down = 4
- left < down
- **Down Left**: down = left
- left > down
- **Left**: left = 4, y = 0
- left > up
- **Up Left**: left = up
- left < up

Now, the reason it was done like this was because I literally built it on top of the previous work I had done - and the `Joy2Rotate` function that existed. In hindsight, we could probably just take the analog value as it is in two axis

```
p1x = ProcessAnalog(DrvAnalogPort0, 0, INPUT_DEADZONE, 0x00, 0xff) / ((256/AnalogRange));
p1y = ProcessAnalog(DrvAnalogPort1, 0, INPUT_DEADZONE, 0x00, 0xff) / ((256/AnalogRange));
```

get rid of `CovertAnalogToRotate()` / your renamed `ProcessAnalogInputs()` and have `Joy2Rotate16()` use the direct `p1x` and `p1y` values to convert them to the 16-direction sectors, and change the way `DrvFakeInput` is used for it, instead of using 4 values just two. It's certainly feasible, and probably if I hadn't been fumbling my way through existing code but rather knew what I was doing, or was now doing it from scratch, that'd probably be a way to go.

The only drawback would be that, in terms of readability, the `Joy2Rotate16()` function would be significantly different to `Joy2Rotate()` in terms of readability/structure, as well as the purpose/usage of `DrvFakeInput`. 

Does this help, was that the question? I probably overlooked something that could achieve this in a much simpler manner :)

Regards.

dinkc64:
The reason I moved the ProcessAnalog stuff to the top of SuperJoy2Rotate() is to get rid of a frame-delay between getting the value and processing it.  It did break downtown, oops :)  thanks for catching that!

I wasn't judging your code, just was curious about that part and trying to better understand it, thanks for explaining!

I did try to find an algorithmic approach to the "x/y axis -> direction", at first I did a pretty bang-up/bad job at generating a matrix and offsetting into the matrix to get the direction.  Then I found some code on google that converts the x/y axis to radians then to direction using a bit of floating point math, but both methods ended up being less elegant than your solution (well, visually/code-wise)

thanks again for explaining things,
best regards,
- dink

pjft:
By all means, thanks for showing interest in it and asking. I assure you that neither I nor my code felt judged, on the contrary - I'm always up for thinking about ideas for making it more compliant with other approaches that have been used in the project, or ways to make it more performant. If you do have suggestions, by all means always bring them up - I have very little attachment to form of the code I write, and I'm always keen on learning and figuring out nice and somewhat readable ways of doing the same thing. :) 

Well done spotting a frame delay between ProcessAnalog and SuperJoy2Rotate(), I completely missed that.

It's been fun looking into this, and I'm happy to have brought a new control method to some of these classic games - and learned about calibr50 in the process :D 

Thank you once again, and keep up the great work on FBNeo :)

Have a good week.

PR #954, opened 2022-03-07
to 2022

taoenwen:


dinkc64:
Hi,
this is regarding d_nes:
for nes_astyanaxc, does it show proper Chinese characters for Released: field for you?

On my side, it looks like this:
game_selector_nes_astyanaxc

I think it's not making proper unicode translation?

best regards,
- dink

PR #962, opened 2022-03-16
CPS1 games (No sound) with QSound [Hack]

taoenwen:
Precompiled instructions are off by default.
Enable by HACK_SOUND=1

### DIFF
--- a/makefile
+++ b/makefile
@@ -53,6 +53,9 @@ endif
 # Include rom set verifying features (comment this for release builds)
 #ROM_VERIFY = 1
 
+# CPS1 games (No sound) with QSound [Hack]
+#HACK_SOUND = 1
+
 # Force recompilation of files that need it (i.e. use __TIME__, __DATE__, SPECIALBUILD).
 FORCE_UPDATE = 1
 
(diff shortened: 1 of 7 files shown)


barbudreadmon:
It only affects hacks ?
Why would we want this disabled ? Those hacks have no sound otherwise, right ?

taoenwen:
For some "No Sound" cps1 games, Get QSound by means of [rom hack],
It only branches in the compile switch selection and doesn't change anything,
The default off is to not affect the real simulation.

dinkc64:
taoenwen, nice! :)

barbudreadmon:
>The default off is to not affect the real simulation.

Hmmm, if there is some interest in having both "simulation" and "qsound version", maybe it could be turned into something that can be chosen at runtime through some dips ?

taoenwen:
@barbudreadmon 
This means that two sound systems must be included,
May cause confusion between [play] and [simulation] ...

dinkc64:
taoenwen, is it ready for merge?

taoenwen:
Yes, it's ready.

PR #963, opened 2022-03-18
IPS address to text

taoenwen:
You can choose to save the IPS file as a text file with the same name to resolve [relative address] and [modified value].
Examples:
0x006c18 0x10
0x00914d 0x05
...

image
image


taoenwen:
Since the time of IPS loading is before ROM decryption,
Its value is not a real modified value,
This function is of little significance.

PR #974, opened 2022-03-28
update sengoku3fs

taoenwen:
Restore the decrypted PROM to the encrypted state to reduce the sum of differences with sengoku3.


taoenwen:
[sengoku3fs.zip](https://github.com/finalburnneo/FBNeo/files/8364544/sengoku3fs.zip)

image

PR #984, opened 2022-04-11
Request for change import from FBN

Taiyoumaru:
I humbly and profusely request that the following change be imported from FBN: 

https://github.com/finalburnneo/FBNeo/pull/716/commits/d0a7b47c2d57ce6febf1e66dbf0bfc6d5ea517fb

This setting has been the source of major frustration for me and many others for more than 15 years of using FB emulators since I and many others use a WASD FGHJ RTYU configuration.

Having the chat button set to T is extremely in the way of actually using it as game inputs. 

Last year, I asked the FBN devs to make it configurable, which was too much, so they agreed to switch it to a button that was not in the way and would be far less likely to be used as game input, since even if you have 2 people playing on the same keyboard, player 2 would most likely use arrows with 6 buttons above or with numpad buttons. 

It is also compatible with all keyboard layouts.

Please please please import this change into FC-FBN, I would be forever grateful!


barbudreadmon:
wtf is that ? closing

PR #985, opened 2022-04-11
Neo Geo: kof2001rpf

Igor-Arabe:
Correct P2 and M1d as on HBMAME.

current version is just hbmame p2 + 1mb of junkdata (bytes00 and bytesFE).
Use the '.bat' to generate junkdata file and join the roms with junkdata file and:
remix pro    - p2 4mb (91668d86) + junkdata = p2 5mb (37a818ed).
original aes - p2 4mb (91eea062) + junkdata = p2 5mb (aa4ef2f0).

[create_junkdata.zip](https://github.com/finalburnneo/FBNeo/files/8465166/create_junkdata.zip)


barbudreadmon:
I'd prefer if we totally replaced this romset by `kof2k1rp` from HBMAME (different name + sprite roms), if the hack is exactly the same

Igor-Arabe:
@barbudreadmon ok. please check.

PR #1000, opened 2022-04-18
nes add some t-chi games

taoenwen:


barbudreadmon:
There are typos, like `Zelda no Densetsu - The Hyrule Fantasy (T-Chi)es` as filename

PR #1001, opened 2022-04-19
nes add some t-chi games

taoenwen:
Thanks！ barbudreadmon

### DIFF
--- a/src/burn/drv/nes/d_nes.cpp
+++ b/src/burn/drv/nes/d_nes.cpp
@@ -15523,6 +15523,210 @@ struct BurnDriver BurnDrvnes_magickidsdorp = {
 	SCREEN_WIDTH, SCREEN_HEIGHT, SCREEN_WIDTH, SCREEN_HEIGHT
 };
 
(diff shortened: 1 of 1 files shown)


barbudreadmon:
i'm kinda confused here, that zelda is not a hack of the original zelda ?

taoenwen:
https://www.ppxclub.com/forum.php?mod=viewthread&tid=711227&extra=&page=1
From the perspective of file size, it should be converted from FDS, and the actual test is mapper 01.
image
Normal operation
image
image
image
image

barbudreadmon:
I believe you are supposed to use "fds_zeldanoden" as parent in that situation, even if they run on different systems, iirc we got a similar situation for some romset on sms/gg. Let's wait for @dinkc64's feedback about this.

dinkc64:
just make it a clone of nes_legenzel

taoenwen:
nes_zeldanodenc clone to nes_legenzel

dinkc64:
:)

PR #1005, opened 2022-04-22
Precompile instructions to dipswitch

taoenwen:
image

### DIFF
--- a/makefile
+++ b/makefile
@@ -53,9 +53,6 @@ endif
 # Include rom set verifying features (comment this for release builds)
 #ROM_VERIFY = 1
 
-# CPS1 games (No sound) with QSound [Hack]
-#HACK_SOUND = 1
-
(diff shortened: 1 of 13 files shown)


barbudreadmon:
Why does it require addendum to our dipswitch parsing logic in `gami.cpp` and `inpdipsw.cpp` ? It can't be handled the standard way ? Like i did in https://github.com/finalburnneo/FBNeo/commit/51aab5f05d343d85c63852e8b2f8091fd562ffce, you create a new dip and simply check its value from within the driver ?

taoenwen:
Yes, check the value of dip and change the driver.
The lightest change is equivalent to adding several sets of drivers to the game.

barbudreadmon:
I still don't understand why you can't handle this dip without changing the logic in `gami.cpp`+`inpdipsw.cpp`.
Well, i can't test this since it'll only work in standalone, let's wait for someone else to review your code.

taoenwen:
That's because I didn't find the case you submitted before, I have to study hard.

PR #1009, opened 2022-04-26
cps-1 : Reserve p-rom length for expanded ips

taoenwen:
[1] Reserve p-rom length for expanded ips (cps-1)

There is no preset length after 68k rom, and an error will be reported when the expanded IPS is encountered.

After judging to be ips patch mode, increasing the length of 68k rom by 0x100000 can solve the problem.

[2] Remove patches that cause binary bloat, which can be implemented with ips (cps-1 / pgm)
[3] update wofjdr to 20220424 (cps-1)


taoenwen:
Test with ips file [prom len is 0x200000]
[wofj.zip](https://github.com/finalburnneo/FBNeo/files/8561634/wofj.zip)

image

PR #1013, opened 2022-04-30
d_neogeo: update kof2k2plus

taoenwen:
Export s1 from c7 & c8 and find that it is consistent with the external s1, so the external s1 is deleted.


taoenwen:
md:
Mega Casanova 3 (HB, v1.0)
md_casanova-04-30-161232

Mega Casanova 2 (HB, v1.0)
md_casanova2-04-30-161315

Mega Casanova 3 (HB, v1.0)
md_casanova3-04-30-161338

Mega Queen Journey, The (HB)
md_queenjour-04-30-161452

taoenwen:
I forgot to check whether I synchronized before submitting.
Now the problem is solved.

PR #1015, opened 2022-05-03
neogeo update

taoenwen:
[1-1] NON-PCB CMC-encrypted games now allow Sprite-ROM to be longer than 0x4000000;
[1-2] The excess also applies to CMC encryption/decryption.
[2-1] In the case of ips, the reserved length of P-ROM/C-ROM/V-ROM is increased.
[2-2] ips extends the length of sprite-ROM, and in this mode, there is a judgment of the true length to ensure that CMC built-in TextRom can be correctly recognized.
[2-3] When synchronizing ips cmc decrypts the length of the temporary memory, it avoids overflows caused by inconsistencies in the length of the two temporary memory.

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -18972,43 +18972,44 @@ struct BurnDriver BurnDrvkf2k2ps2 = {
 };
 
 // The King of Fighters 2002 (PlayStation 2, Hack)
-// Hack by Dream
+// Hacked by Dream
 
-static struct BurnRomInfo kf2k2ps2bRomDesc[] = {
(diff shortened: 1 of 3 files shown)


taoenwen:
Test: 
-------------------------------------
[1] ips with increased capacity
The host is: kof2002 / kof97
[kof2002.zip](https://github.com/finalburnneo/FBNeo/files/8609328/kof2002.zip)
[kof97.zip](https://github.com/finalburnneo/FBNeo/files/8609332/kof97.zip)

[2] Sets for increasing capacity
https://drive.google.com/file/d/10imeV0rEvkSZM2BGy09m0kScK8zS_efJ/view?usp=sharing

[3] binary file
[fbneo.zip](https://github.com/finalburnneo/FBNeo/files/8609390/fbneo.zip)

PR #1019, opened 2022-05-04
update d_vendetta.cpp :

taoenwen:
sync mame : added PLDs (BAD_DUMP / Bruteforced)


dinkc64:
taoenwen, ym2151 volume too low now (0.05), please check.

taoenwen:
@dinkc64 Yes, my mistake!

image

dinkc64:
please keep in mind: we set a lot of volumes different from mame on purpose, so please make sure it really sounds better or more balanced if you change a volume, instead of changing it just because mame did.

best regards,
- dink

taoenwen:
I understand

taoenwen:
@dinkc64 Thanks!

PR #1027, opened 2022-05-09
update cps-1 / neogeo / nes

taoenwen:
[cps-1] update wofjdr to 20220507
[neo_run] Remove unused variable
[nes] add / update T-Chi games


dinkc64:
thanks taoenwen, very nice work! :)

PR #1029, opened 2022-05-12
Reduce latency by adjusting emulated frame slice boundary

energy-t:
finally got around to fixing this issue that i previously described on the libretro forums over here: https://forums.libretro.com/t/strange-behavior-in-mame-lag-control/35082/43

this patch changes the order of emulated scanilnes from [0-264) to [248, 264) followed by [0, 248). this reduces latency for any games that poll input shortly after vblank as well as reduces the variance of response times for games that oscillate between emulated frame slices under the current configuration, as the metal slug games do. ex: metal slug 2/x will now always (provided the game is not undergoing slowdown) show a response when frame stepping the emulation within 3 to 4 frames rather than 3 to 5 frames

i have another branch with the same fix that can be toggled on/off with a #define variable along with printing some debug info that i used to help test to make sure everything was working as it should, which you can find here: https://github.com/energy-t/FBNeo/tree/neogeo-reduce-latency-test

### DIFF
--- a/src/burn/drv/neogeo/neo_run.cpp
+++ b/src/burn/drv/neogeo/neo_run.cpp
@@ -342,6 +342,16 @@ static INT32 nff0002 = 0;
 static INT32 nff0004 = 0;
 
 
+static inline INT32 NeoCurrentScanline()
+{
+	return (SekCurrentScanline() + 248) % 264;
+}
+
+static inline INT32 NeoCurrentScanlineOffset()
+{
+	return SekTotalCycles() - SekCurrentScanline() * nSekCyclesScanline;
+}
+
 bool IsNeoGeoCD() {
 	//return (nNeoSystemType & NEO_SYS_CD);
 	return ((BurnDrvGetHardwareCode() & HARDWARE_PUBLIC_MASK) == HARDWARE_SNK_NEOCD);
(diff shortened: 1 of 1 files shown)


barbudreadmon:
Thanks for your contribution !

I remember talking with @dinkc64 about this but i suppose it totally slipped out of our mind afterward, sorry about that.
Did you take care of the raster issues ?

energy-t:
yep, did indeed!  at least i believe so.  first thing i tested it with was neo turf masters like u advised b4 and the raster effects are fully preserved now.  the fix for that problem was quite simple just grokking the code to get to that point, not so much..  anyway, i used the test branch i linked to that spits out exactly where in a frame a game polls input in order to make sure things lined up properly before and after the adjustment on a bunch of games and it all seems to check out now

barbudreadmon:
Ok, i'm not noticing any regression and we can always fix it later if anything show up, merging !

dinkc64:
very cool!  how's the neocd stuff hold up?

barbudreadmon:
The few neocd titles i tried seemed to work fine too

energy-t:
@barbudreadmon or @dinkc64 : very sry for the inconvenience but i have a small bugfix for this over here: https://github.com/energy-t/FBNeo/tree/neogeo-reduce-latency-fix

this fixes an issue that can occur in the metal slug games (particularly 1) when firing the shotgun.  the blast is supposed to white out all sprites for 2 frames but after the patch it will sometimes white out only some sprites on 1 frame, all on the 2nd, and then rest on the 3rd.  you get a shotgun at about 40% into stage 2 after destroying one of the boats if you want to see for yourself.

anyway, delaying the renderer updating the palette to just before any visible scanlines fixes this.  i think technically this or similar issues could've also occured prior to the frame boundary adjustment patch but the patch just made it far far more common.

should i create a new pull request?  not sure what would be best..

dinkc64:
Hi,
Sure, please make a PR

it's not an inconvenience, it's progress :)  

best regards,
- dink

PR #1034, opened 2022-05-15
neogeo: bugfix for minor gfx issue introduced in commit - f412f7b5387…

energy-t:
…ae04fcda42a0b00321cb2c051ff0e

fixes an issue that can occur in the metal slug games (particularly 1) when firing the shotgun. the blast is supposed to white out all sprites for 2 frames but after the prev patch it will sometimes white out only some sprites on 1 frame, all on the 2nd, and then rest on the 3rd.  delaying the renderer updating the palette to just before any visible scanlines fixes this

### DIFF
--- a/src/burn/drv/neogeo/neo_run.cpp
+++ b/src/burn/drv/neogeo/neo_run.cpp
@@ -4805,12 +4805,6 @@ INT32 NeoFrame()
 	bRenderImage = false;
 	bForceUpdateOnStatusRead = false;
 
-	if (pBurnDraw) {
-		NeoUpdatePalette();											// Update the palette
-		NeoClearScreen();
-	}
-	nSliceEnd = 0x10;
-
 	SekNewFrame();
 	ZetNewFrame();
 
@@ -4875,6 +4869,11 @@ INT32 NeoFrame()
 	bForcePartialRender = false;
 
 	// Display starts here
+	if (pBurnDraw) {
+		NeoUpdatePalette();											// Update the palette
+		NeoClearScreen();
+	}
+	nSliceEnd = 0x10;
 
 	nCyclesVBlank = nCyclesTotal[0]; //nSekCyclesScanline * 248;
 	if (bRenderLineByLine) {


dinkc64:
Hi energy-t,
Just got a report on the forum that Blazing Star's second level bg graphics are flickering, could you please check into it?

best regards,
- dink

energy-t:
will do

barbudreadmon:
Maybe the raster line by line thingy i enabled for cross sword 2 the other day is also needed here ? Flickering already existed in cs2, but it was somehow more noticeable after energy's changes.

energy-t:
> Maybe the raster line by line thingy i enabled for cross sword 2 the other day is also needed here ?

^ ya that was the first thing i checked too, didn't help unfortunately.  i'm still working on this, i have few ideas where the problem may lie but i'm going to have to write some scaffolding debug code to help me out.. in the mean time i've created a new branch that provides the ability to completely restore the driver's old behavior via global static bool that can be set on a per game basis (with just blazing star set to use it atm) over here:

https://github.com/energy-t/FBNeo/tree/neogeo-shift-frame-boundary
https://github.com/energy-t/FBNeo/commit/9dac6416630f43b3c617158066e65bce195f4613

lmk if that would be acceptable for the time being and i'll go ahead and pull request it.  it would put my mind at ease a bit especially since i'm not very well versed in using git and it's all been a bit overwhelming for me rn

energy-t:
ok i found the problem so ignore that last comment, i'm going to try and get the fix done and pull request in before i go to sleep but if not i'll have it ready within next 24 hrs for sure

dinkc64:
thanks :)

dinkc64:
energy-t, re-recording got broken somehow, I'll try to figure out the problem, but if you have any ideas before I fix it, let me know :)

re-recording is a TAS feature where you can record inputs (Game -> Replay -> Record Input) and save/load states during the recording, it depends on really tight cycle timing.  It shows up pretty quickly with a game of kof94.

dinkc64:
sorry about the accusatory message re: rerecording, I made a mistake copying back the old neo_run.cpp and re-compiling, and it lead me to falsely believe it was related to your code.  It actually exists prior to your code as well, but just with the game kof94 (so far).  When I finally came to realization of mistake, I just deleted the message and went to sleep like an idiot.  On the other hand, if you notice anything that would cause kof94 to lose determinism, please let me know :)  I have a feeling it might be with the upd timer chip, but I havn't explored it yet.


Regarding cs2, that game had always been very glitchy graphically, so I wouldn't worry about it unless another broken game comes up (so far so good)  :)

best regards,
- dink

dinkc64:
kof94 and perhaps all ng games should enjoy happy re-recording now :)

energy-t:
awesome, glad it got sorted out!  sorry i wasn't able to be of any help.  i was going to try and at least take a look at this but have been super busy w/life stuff.  though looking at the patch set i can tell it's not something i would've been able to spot w/o good deal of work / studying fbneo's neogeo driver past what i'm already familiar with (which isn't much)

PR #1051, opened 2022-05-28
d_brkthru.cpp : add brkthrut & brkthrubl

taoenwen:
sync mame
keep the format consistent

brkthrut:

brkthrubl:


taoenwen:
test sets: same as mame
[brkthrubl.zip](https://github.com/finalburnneo/FBNeo/files/8790986/brkthrubl.zip)
[brkthrut.zip](https://github.com/finalburnneo/FBNeo/files/8790987/brkthrut.zip)

PR #1058, opened 2022-06-04
d_neogeo.cpp : test

taoenwen:
Simplified operation of the third P-ROM


taoenwen:
@dinkc64 @barbudreadmon 
Hi dink， Hi barbudreadmon
I have some confusion about the third P-ROM:
[1] Ideally, the third P-ROM is not involved in loading for the first time.  (use : 0 | BRF_ESS | BRF_PRG)
[2] Then by loading into temporary memory, map to 0x900000 ~ 0x900000 + [size]
[3] Destroy temporary memory on exit

I've been thinking that it doesn't seem necessary to be that complicated if it's just for mapping memory addresses
Am I wrong? Also please correct me, thank you very much!

taoenwen:
If feasible, a macro template, can solve these problems

dinkc64:
It's a nice simplification :)

PR #1059, opened 2022-06-05
Rev B of the ROM

mobasuite:
Thanks for reading this, unsure which rev was the OG included in this, but this is the REV B either way.

With Header : 4E 45 53 1A 20 00 E3 10 00 00 00 00 00 00 00 00

### DIFF
--- a/src/burn/drv/nes/d_nes.cpp
+++ b/src/burn/drv/nes/d_nes.cpp
@@ -21294,7 +21294,7 @@ struct BurnDriver BurnDrvnes_nemulesis = {
 };
 
 static struct BurnRomInfo nes_pixelpoopsRomDesc[] = {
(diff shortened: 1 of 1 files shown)


dinkc64:
is that the original header that came with the rom?  if not, can you share the original unmodified header?

mobasuite:
That's what I got from my forum which was dated by the no intro group.

mobasuite:
Here's the file CRC anyway : 

CRC32: | 17812952
MD5: | ffa1023dfa8f887a98d7debb5eab8d09
SHA-1: | 671d0ec8ae8e73a1eaa465687a2c38e0e0747ed4

mobasuite:
NES ROMs don't have watermarks

dinkc64:
thanks
thanks for the extra info Igor-Arabe, one can only wonder what those 4 bytes are for :)

PR #1066, opened 2022-06-11
(luaengine.cpp,m68kcpu,m68000_intf) - fix luamemhook fns

NBeing:
I can't figure out why the CALLBACK_PC_CHANGED  callback is not exposed to lua when working this way. It seems to never get updated!

Getting the Z80 to work has similar problems as m68k, It seems the way it was handled was changed as well since fba-rr.

Let me know what errors I have made, and I will try my best to fix them.

The changes have been tested by Alice and myself from the LUA scripting end.

### DIFF
--- a/src/burner/luaengine.cpp
+++ b/src/burner/luaengine.cpp
@@ -1161,29 +1161,6 @@ struct registerPointerMap
 
 #define RPM_ENTRY(name,var) {name, (unsigned int*)&var, sizeof(var)},
 
-registerPointerMap m68000RegPointerMap [] = {
-	/*
-	RPM_ENTRY("pc", M68000_regs.pc)
-	RPM_ENTRY("d0", M68000_regs.d[0])
-	RPM_ENTRY("d1", M68000_regs.d[1])
-	RPM_ENTRY("d2", M68000_regs.d[2])
-	RPM_ENTRY("d3", M68000_regs.d[3])
-	RPM_ENTRY("d4", M68000_regs.d[4])
-	RPM_ENTRY("d5", M68000_regs.d[5])
-	RPM_ENTRY("d6", M68000_regs.d[6])
-	RPM_ENTRY("d7", M68000_regs.d[7])
-	RPM_ENTRY("a0", M68000_regs.a[0])
-	RPM_ENTRY("a1", M68000_regs.a[1])
-	RPM_ENTRY("a2", M68000_regs.a[2])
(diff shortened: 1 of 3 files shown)


dinkc64:
hmm, I'm going to have to revert this for now, because it breaks all of the non-windows builds, you see - only the windows version uses lua, so the m68k shouldn't be using any lua functions, or at least we'll have to ifdef them for the windows version.

NBeing:
Understood! I wasn't even aware a *nix build existed. 

I think I cannot modify this PR unless I'm mistaken, but I have made a sample here:
https://github.com/NBeing/FBNeo/pull/3
If this looks more accurate, I can reopen a new PR from that commit.

dinkc64:
Everything looks good, except for a little suggestion:
	#include <m68k/m68kcpu.h>  <- remove this in m68000_intf.h, then put the function decl. for m68k_set_pc_changed_callback in m68k/m68k.h - search for "dink" and put it near there :) 

best regards,
- dink

NBeing:
Hello Dink,

Thanks so much for being patient with me and walking me through this.

How is this looking?

https://github.com/NBeing/FBNeo/pull/3

Best,

N-Bee

dinkc64:
Hi N-Being,
I'm always glad to help out :)
On second thought, could you please remove this from m68k.h:
/* For m68k lua implementation */
#if defined (BUILD_WIN32)
static void (*pc_changed_cb)(UINT32);
#endif	

Then put the "static void (*pc...." back in m68kcpu.c where it was, sorry for the confusion!  If everything works fine after that, please submit the pr.

best regards,
- dink

NBeing:
Hello Dink,

I have hopefully reflected your suggestions accurately:
https://github.com/NBeing/FBNeo/pull/3/files#diff-bfce72c7dc1d6dfe29a8e80a5fbdfb75a2cf48340c61b622b069add70eabddac

Best,
N-Bee

taoenwen:
@NBeing Hi，NBeing，Give a suggestion, [#pragma once] might not be suitable for gcc
Maybe using this way will improve compatibility
`#ifndef xxx
#define xxx
// to do
#endif`

NBeing:
@taoenwen Thank you, those two files were committed by accident! Will correct. Thank you for your eyes.

NBeing:
closing this and reopening here: https://github.com/finalburnneo/FBNeo/pull/1072

PR #1082, opened 2022-06-23
SDL2: add a simple controller buttons mapping menu

cjom:
It is now possible to map joysticks buttons or remap game controllers buttons in ingame menu.

A file named gamecontrollerdb.txt is created in config folder (where fbneo.ini is) and loaded when fbneo starts.

It is possible to manually create or edit gamecontrollerdb.txt (for example, to change controller name).


tmaul:
Nice work :)

PR #1099, opened 2022-07-06
SDL2: fix 4 players support

cjom:
When using keyboard only a controller for player 2 was properly set.

When not using keyboard, controller 3 and 4 would not be properly mapped.

Note about function GameInpConfigOne: seems to have cases that are not called in any way (X-Arcade and HotRod).


dinkc64:
I think the x-arcade and hotrod were popular arcade-like controllers for pc a long time ago, but probably not relevant now.  who knows? :)

PR #1102, opened 2022-07-08
mslug4a : Forced display difficulty [Alice愛麗絲]

taoenwen:


barbudreadmon:
imho, it's a huge workaround for avoiding 1 manipulation in 1 hack

taoenwen:
I think so too.

PR #1105, opened 2022-07-13
SDL2: fix inputs from <romname>.ini being ignored (and presets/*.ini …

cjom:
…too)

Any changes made to inputs would always be ignored and reset to defaults.

Also fixed presets being ignored in SDL builds (for example, presets/cps.ini will be used for any CPS game if <romname>.ini did not exist).

Note about -joy parameter and player 1 using keyboard or joystick:
If <romname>.ini or presets/*.ini exists when a game is launched, whatever is defined for player 1 is what is used, being irrelevant if -joy parameter is used or not.

### DIFF
--- a/src/burner/burner.h
+++ b/src/burner/burner.h
@@ -136,8 +136,9 @@ void GetHistoryDatHardwareToken(char *to_string);
 extern INT32 nAutoFireRate;
 
 // Player Default Controls
-extern INT32 nPlayerDefaultControls[8];
-extern TCHAR szPlayerDefaultIni[5][MAX_PATH];
+#define MAX_JOYSTICKS 4
+extern INT32 nPlayerDefaultControls[MAX_JOYSTICKS];
+extern TCHAR szPlayerDefaultIni[MAX_JOYSTICKS][MAX_PATH];
 
 // mappable System Macros for the Input Dialogue
 extern UINT8 macroSystemPause;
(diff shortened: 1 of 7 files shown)


dinkc64:
Do you think you could localize gami.cpp in the sdl branch and leave the one in burner alone?
I'm wary about these changes breaking things in the Windows UI

best regards,
- dink

cjom:
> could localize gami.cpp in the sdl branch
Sure, but I need some pointers on how to do it... I still don't see how to put a file in "sdl" folder and use same name functions without conflicting with shared files functions...
I mean, are they declared somewhere?

cjom:
In the mean time, I compiled for windows and played a bit, tested inputs and presets from 1 to 4 players.

I found no issues, but I got a better understanding of the X-arcade thing and agree it's better leave that part untouched, although it seems this way 4 players can't all use gamepads.

But the ConfigGameLoadHardwareDefaults still needs to be changed for SDL to find and load presets, but because it's inside "#if defined" I suppose it is acceptable as it won't change windows part...

dinkc64:
Just copy the file into the sdl folder as gami_sdl.cpp, and have your makefile compile it instead of burner/gami.cpp, should be ok :)

cjom:
Thanks for the clarification. Suspecting something like that I changed the code again to leave that part as it was.

I have already reverted everything in gami.cpp, only change needed it's related to path of ini files,  it's a change similar to what is done in other files with " #if defined(BUILD_SDL2) && !defined(SDL_WINDOWS)    etc "

I think we can close this pull,  I'll test more and make a new pull request tomorrow.

dinkc64:
Alright, sounds good & looking forward to it :)

PR #1122, opened 2022-07-23
I wish fbneo had the same 64Mhz CPU limit as WinKawaks™, Is it possible ?

alicemsu:
I don't know if there is actually some error with the clock values but the only way I was able to achieve 533% was this....
[numdial.txt](https://github.com/finalburnneo/FBNeo/files/9174770/numdial.txt)


barbudreadmon:
what's wrong with people opening random PR instead of issues ? is there some kind of github bug or something ?

alicemsu:
IDK maybe, I'm new to using this, so I hope my suggestion has arrived.

PR #1123, opened 2022-07-24
Add konami gx games

juan200213:
Hello
In mame there games of konami gx like fantastic journy and lethal enforcers 2 please add this games thanks


dinkc64:
THIS IS NOT A PLACE FOR REQUESTS, THANKS

juan200213:
So where I put request

barbudreadmon:
@juan200213 nowhere, we don't accept requests, **this is written in the rules you are supposed to read before opening issues**

PR #1125, opened 2022-07-26
d_ssv: save NVRAM in states for games that have it mapped

KScl:
Primarily to fix Change Air Blade (`cairblad`), because it does not like it if NVRAM initialization gets skipped somehow.


barbudreadmon:
Thanks for your contribution !

PR #1128, opened 2022-07-29
d_cv1k: first attempt at reducing runahead memory footprint

barbudreadmon:
@dinkc64 cv1k runahead gets fairly playable on my computer with those changes (about 5 times faster), and i didn't notice any emulation issue, does it look safe to you ?

### DIFF
--- a/src/burn/devices/epic12.cpp
+++ b/src/burn/devices/epic12.cpp
@@ -999,7 +999,9 @@ void epic12_scan(INT32 nAction, INT32 *pnMin)
 	SCAN_VAR(epic12_device_blit_delay);
 	SCAN_VAR(m_delay_scale);
 	SCAN_VAR(m_blitter_busy);
-	ScanVar(m_bitmaps, m_gfx_size * 4, "epic12 vram");
+	if (~nAction & ACB_RUNAHEAD) {
+		ScanVar(m_bitmaps, m_gfx_size * 4, "epic12 vram");
+	}
 	thready.scan();
 }
 


Tatsuya79:
Wow pretty cool.

Just tested saidaioujou with the libretro core:
330fps normal
~80fps min runahead 2 frames

dinkc64:
the cache and the vram in epic12 will be a problem, though, I may be able to reduce the cache to a very small amount.
the one in epic12 (bitmaps) is the vram + screen, so after it goes from the runahead frame (+2 frames) back 2 frames, it'll basically still have the screen from 2 frames ahead.  so with runahead, the video will have artefacts from the "future" (+2frames), which might or might not be apparent.  Do some framestepping on scenes with lots of change to see if you notice anything?

Tatsuya79:
I tried frame stepping while changing direction in saidaioujou, akai katana, mushi futari.
Can't see anything particular; the shadows are following the ships, the explosions animation are fine, the sprites on ground follow the ship.

dinkc64:
curiously, does lr have an accurate way to tell if state is runahead or normal now?

barbudreadmon:
>curiously, does lr have an accurate way to tell if state is runahead or normal now?

Yes, i ended up implementing it myself last month.

dinkc64:
barbudreadmon, sorry for keeping you so long on this, but I was pretty much overwhelmed by the rewind project... which lead to finding out more core stuff needed rewriting/intense fixing.  a real can-o-worms it turned out to be! :)
Could you meet me around your morning (before 1:30 detroit time) or my morning (9am detroit time)?  I have something nice for you.

PR #1144, opened 2022-08-20
fix sdl windows build

MistyDreams:
@cjom getline is posix so ive added the function for windows as its missing. The only other alternative is to use c++ getline. The other fix is to disable unicode for all windows builds accept the original fork that uses it.

### DIFF
--- a/src/burn/devices/thready.h
+++ b/src/burn/devices/thready.h
@@ -10,7 +10,7 @@
 
 #if defined(WIN32)
 #define THREADY THREADY_WINDOWS
-#if !defined(UNICODE) && !defined(__LIBRETRO__)
+#if !defined(UNICODE) && defined(BUILD_WIN32)
 #define UNICODE
 #endif
 #include "windows.h"
--- a/src/burner/sdl/sdl2_gui_ingame.cpp
+++ b/src/burner/sdl/sdl2_gui_ingame.cpp
@@ -56,6 +56,75 @@ static UINT16 current_menu = MAINMENU;
 static UINT16 current_selected_item = 0;
 UINT16 firstMenuLine = 0;
 UINT16 maxLinesMenu = 16;
+//getline is not avaiable on windows lets emulate it
+#if defined(WIN32)
+
+#include <stdlib.h>
+#include <errno.h>
+#include <limits.h>
+
+ssize_t getdelim(char **lineptr, size_t *n, int delim, FILE *stream) {
+    char *cur_pos, *new_lineptr;
+    size_t new_lineptr_len;
+    int c;
+
+    if (lineptr == NULL || n == NULL || stream == NULL) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    if (*lineptr == NULL) {
+        *n = 128; /* init len */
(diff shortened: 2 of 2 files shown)


dinkc64:
Where is this getline code copy and pasted from?

MistyDreams:
https://github.com/ivanrad/getline

https://github.com/ivanrad/getline/blob/master/UNLICENSE

dinkc64:
great, nice find :)

dinkc64:
please remove the change to thready, barbudreadmon beat you to it

MistyDreams:
its change for his his only covers libretro

dinkc64:
thanks for clarification

MistyDreams:
no problems pasted a reply in a wrong link sorry about that deleted it

cjom:
@MistyDreams still about "getline"... I noticed now that using this function brings even more different approaches to an already "not uniform" app... other functions are using "_fgetts" that I copied from gami.cpp, also config.cpp just uses "fgets" to rad from *.ini files.

Wouldn't be better not use getline at all and try to use always the same function?
And would it be _fgetts? Some google result mentions is a Windows function...?

MistyDreams:
I dont  know what the general approach is here is. I Just fixed the windows sdl build to include the function that windows didnt have that it was using already. Since its multi platform you would need to tie in with the admins here.

cjom:
Well, since SDL version is using "char" instead of "TCHAR" both saving/loading configs, I suppose it's more efficient to keep using fgets (as it is already used there) instead of a different function.
I will investigate...

MistyDreams:
I managed to get the cross compile working properly for mingw with with a few makefile changes so can test windows easier now. I had to add some debugging to test I think it might be handy keeping the debugging stuff in though.

```
INEPATH=/usr/i686-w64-mingw32/bin ./fbneo.exe rygar                                                                                                     ✔  4m 36s  
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
007c:fixme:hid:handle_IRP_MN_QUERY_ID Unhandled type 00000005
FBNeo v1.0.0.03
INFO: We compiled against SDL version 2.24.0 ...
INFO: But we are linking against SDL version 2.24.0.
INFO: SDL Audio driver(s) 0: wasapi
INFO: SDL Audio driver(s) 1: directsound
INFO: SDL Audio driver(s) 2: winmm
INFO: SDL Audio driver(s) 3: disk
INFO: SDL Audio driver(s) 4: dummy
INFO: SDL Video driver(s) 0: windows
INFO: SDL Video driver(s) 1: dummy
0024:fixme:system:QueryDisplayConfig (00000002 0598ED74 06936A00 0598ED78 06936A50 00000000): semi-stub
0024:fixme:system:DisplayConfigGetDeviceInfo DISPLAYCONFIG_DEVICE_INFO_GET_TARGET_NAME: stub
SDLSoundInit (44100Hz) (6000FPS)
```

PR #1147, opened 2022-08-25
fix cross compile for sdl2

MistyDreams:
Fix cross compile for mingw. Will post details of why this is needed.

```
#set env flags for mingw
. /usr/bin/mingw-env x86_64-w64-mingw32  

 [FBNeo]$ pkgconf --libs sdl2
-L/usr/x86_64-w64-mingw32/lib -lmingw32 -lSDL2main -lSDL2 -mwindows 
[FBNeo]$ sdl2-config --libs 
-L/usr/lib -pthread -lSDL2
```
I really only have a laptop with windows on it so for testing so i need to cross compile. There is an issue with sound which doesnt happen in wine and does in windows. I will debug that later today hopefully to get all sound drivers working in sdl. Currently only direct sound is working and its forced anyway so not really a immediate must fix issue.

### DIFF
--- a/makefile.sdl2
+++ b/makefile.sdl2
@@ -11,10 +11,13 @@ ifeq ($(UNAME_S),Darwin)
 DARWIN=1
 endif
 
+LD=$(CC) 
+
 
 ifeq ($(OS),Windows_NT)
 WINDOWS=1
 endif
+
 
 #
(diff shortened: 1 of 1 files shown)


dinkc64:
This broke the auto build process for sdl2

MistyDreams:
what is the error ?

PR #1150, opened 2022-08-28
Cross compile

MistyDreams:
Added a fallback to sdl2-config if pkgconf fails. I dont have any details of the fails but if this fails can you drop me some logs so I can look into it.


MistyDreams:
Just checked some information on this could be called pkgconf or pkg-config will add the pkg-config and test. Dont pull this for until I update this and test so we dont end up with another pull request.

MistyDreams:
tested on linux, with and without cross compile. I also tested on msys2. Checks all varieties of sdl2-config  and lets you know if sdl2 isint installed. Is good to go. If there is any issues let me know. The only error from the changes before would be pkgconf is not available where you build your releases. It checks all 3 pkgconf, pkg-config and sdl2-config if it finds none it will let you know sdl2 isint installed.

PR #1153, opened 2022-08-29
Update Waldo's Life to current version

mobasuite:
There is now EN and one ES

The size was -20kb for both

Not sure why github is displaying as 9000 deleted 9000 added...


dinkc64:
How to fix your problem - 2 solutions, though, it's best to set both.

1: set git to auto-convert line endings to project's style
... https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings
2: set your editor to preserve the line endings/format of original file

PR #1178, opened 2022-09-28
fix nested if and tidy sanatize up

MistyDreams:
SYMBOL was nested between sanitize and profile definitions. sanatize will  also automatically set symbols if not set for easier debugging,


dinkc64:
MistyDreams closed this by deleting the head repository [7 hours ago]

This is scary, can you re-do the pr without that? :)

PR #1189, opened 2022-10-03
update msvc project for mcr68

taoenwen:


dinkc64:
Thank's taoenwen ! :)

PR #1216, opened 2022-10-21
Updated NES ROM

B94:
Removed clone property of skartfighter and updated the CRC


B94:
This was not the branch I wanted. Sorry.

PR #1217, opened 2022-10-21
d_sailormn: fix memory request for sailormnffj

taoenwen:


barbudreadmon:
I wasn't totally happy with this PR, because the code might get messy if new hacks using different sizes are added, so i went with https://github.com/finalburnneo/FBNeo/commit/2cb773b0edfad1dc21d678abb51e29e8b4c04916 instead. You might want to do another PR for ips support.

taoenwen:
Yes, you've thought it through quite well and it's wonderful, thanks!

PR #1228, opened 2022-11-09
d_megadrive: md_megamanfix

Igor-Arabe:
Added correct crc32, romname.

base: megamanrb rom + (U-RB)-v105 ips

I believe the original intention was add jp-rom and megamanrb with  "fixer upper" patch. if this isn't correct let me know.


Igor-Arabe:
Now is ok.

dinkc64:
I think gab75 has updated this already, but not sure ?

Igor-Arabe:
Not completely. He added megamanfix with crc and romname like as on megamanrb.

check commit: https://github.com/finalburnneo/FBNeo/commit/6ce192221edc106a327d3ce72f7165359bb77a4c

dinkc64:
oh, I see - thanks!

PR #1229, opened 2022-11-10
sync mame

taoenwen:


dinkc64:
@taoenwen very nice.  I ported hap's fixed sound for olibochu last night, will sync. to git soon!  With the palette fixes that you've added it's very nice.

best regards,
- dink

PR #1243, opened 2022-11-22
Native Windows 11 SDK

mobasuite:
Might require some changes to Github Actions.


mobasuite:
It would possibly increase performance on windows systems, for the small audience that builds it themselves.

PR #1244, opened 2022-11-23
d_nes: skartfight

Igor-Arabe:
"NULL" isn't parent rom. Added correct name.


mobasuite:
fixed, close

dinkc64:
Thanks Igor-Arabe, originally I just copied this from fightcade repo

PR #1261, opened 2022-12-08
update the full names of some cave games

taoenwen:


taoenwen:
(d_donpachi.cpp) I referenced mame's corresponding naming in cave.cpp, maybe I should have left it as it was (for consistency)

taoenwen:
@dinkc64 It's been fixed.

PR #1313, opened 2023-01-21
win32: fix neogeo ghost issue in msvc x86

taoenwen:


taoenwen:
image

Working well now

PR #1325, opened 2023-01-29
d_gladiatr: fix a coin issue

taoenwen:
d_gladiatr: add Ougon no Shiro (Japan)


dinkc64:
thanks for fixing my bug :)

PR #1432, opened 2023-04-08
Added hiscore support to multiple games

shaunquick:


shaunquick:
I think I have added hiscore support to most that were missing - there may be a few that I overlooked, I'll look at comparing hiscore.dat with which games have not had hiscore support for which ones that need to be uplifted - 
Also note - when I added hiscore support to pipibibsp, it was successful.
For the games pipibibs & pipibibsa, they work without hiscore support but do not work with it, so I have not added hiscore support to these.

PR #1436, opened 2023-04-11
added hiscore support to multiple games

shaunquick:
Added krull, beaminv and pacominv

So there are now only 3 games that are in hiscore.dat that cannot be supported by the standard BDF_HISCORE_SUPPORTED code - not sure what the issue is but here are the three games.

amspdwy - breaks game when hiscore support added
pipibibs - breaks game when hiscore support  added
pipibibsa - breaks game when hiscore support  added


dinkc64:
hi shaunquick,
thanks for adding all those hiscores! :)
I've committed a fix for amspdwy, will try to get a fix in for the rest later tonight!

best regards,
- dink

dinkc64:
regarding pipibibs:
I can't get it to fail on my side after enabling HS, any suggestions? :)

best regards,
- dink

shaunquick:
Hi there - I cannot make it fail - thank you for the fix !

I have created a python script to identify any games that have missing hiscore support (happy to share, works on windows) that are BDF_GAME_WORKING and can report that there are no more left at this time :-)

dinkc64:
wow, great news! :)
Thanks for all the help!  I'm sure the people / friends & fans of fbneo will appreciate it just as well!

best regards,
- dink

shaunquick:
found a few more games (after I downloaded most up to date hiscore.dat).
Games that break when I add hiscore support are -

robotron
robotron87
robotrontd
robotron12
robotronun
robotronyo

I have not updated these games.
I suspect that there is a memory problem but cannot work out what it is.

dinkc64:
hmm, I've grabbed the latest hiscore from leezer's site, and with it, it won't write robotron.hi (after I've added BDF_HISCORE... to robotron..) - any ideas? :)

best regards,
- dink

PR #1458, opened 2023-05-01
Auto Reload after some DIPs are selected

taoenwen:
aud_dsound3.cpp : Prevents being changed to NULL externally while looping
cps1/pgm/neogeo : Auto Reload after some DIPs are selected ( qsound / old code / ver )


taoenwen:
mslugxdg test : https://drive.google.com/drive/folders/1qqk2cNlE5ecb6wMGjJmAqTef2xVgshyJ?usp=share_link

dinkc64:
Hi taoenwen,
I think it's really not a good idea to call DrvInit from inside frame.  I'll try to think of a better way to do it..  Hmm.

taoenwen:
OK

PR #1517, opened 2023-07-10
pre90s, pst90s fixes. samsho2pe update

Igor-Arabe:
- hatrisp: correct size/crc
- terrafua: correct sizes
- samsho2pe: updated to version (23-07-10)


dinkc64:
thanks! :)

PR #1553, opened 2023-08-30
win32: add cmdline for romdata

taoenwen:


dinkc64:
Just a note:
memset() only sets 1 byte, _T("\0") is 2 bytes
if you have something like
TCHAR szBuf[100];
use this:
memset(szBuf, '\0', sizeof(szBuf));

using sizeof() is a good line of protection in-case the structure size changes

best regards,
- dink

taoenwen:
Thank you for your guidance.
I'll fix it.

dinkc64:
Hi taoenwen,
memset() and malloc() operate @ byte-size, so it is still necessary to specify the size * sizeof(wchar_t) for, example:
pszFullName[i] = (wchar_t*)malloc(MAX_PATH * sizeof(wchar_t));
..
memset(pszFullName[i], '\0', MAX_PATH * sizeof(wchar_t));

best regards,
- dink

taoenwen:
I see. Thank you for taking the time to explain it to me.

taoenwen:
The use of memset() has been fully adjusted
[1] String arrays using sizeof()
[2] malloc the actual length of the requested string.

taoenwen:
yes，sir

PR #1568, opened 2023-09-22
d_neogeo: samsho2pe

Igor-Arabe:
latest version 2023-Sep-22:
- Randomize Collor in Roulette Char.
- Randomize Maps in Versus Mode, Not Versus CPU.
- Remove combo 100% Nicotine, bug fix.
- Add 8 frames in recovery for special grab for galford and hanzo, similar to a kof (example, clark, ralf).


dinkc64:
remove 100% Nicotine, :(

PR #1584, opened 2023-10-15
SDL: Fix debug print function

12Me21:
In the SDL/SDL2 port, `AppDebugPrintf()` (and thus `bprintf()`) is completely broken:
It passes a `va_list` to `printf` (!!) (instead of `vprintf`), causing all debug prints to output garbage data & occasionally segfaults.
(In particular, i was getting crashes whenever hiscore data was written to memory, because one of the debug messages happened to trigger a segfault)


barbudreadmon:
Thanks !

PR #1595, opened 2023-11-11
burn.cpp: fix incomplete string copy

taoenwen:


dinkc64:
Hi,
can you show me what you're doing to replicate the issue?  I'm asking because wcscpy() does copy the terminating null, but in cases when it doesn't, it might be a far more serious issue (OOB etc)

best regards,
- dink

taoenwen:
image
L"Guardians\0\u96FB\u795E\u9B54\u584A \uFF29\uFF29\0" It will stop after "Guardians".

dinkc64:
oh, I see - nevermind then :)

barbudreadmon:
How come there is a `\0` in the middle of this description ?

I've seen quite a few of them too in the non-unicode descriptions, and replaced them since they simply didn't seem to work in the contexts where `BurnDrvGetTextA(DRV_FULLNAME)` is called.

dinkc64:
barbudreadmon, it puts a little "dot" between the ascii name and non-ascii name, see:
(feature?)

gamelist

dinkc64:
@taoenwen, did you notice all the stray CR (\r or 0x0d) in the code?

stray_cr

taoenwen:
@dinkc64 I can't see the difference between them in the VS editor, so it looks like I should use LF to solve them

taoenwen:
I'll fix them right away.

taoenwen:
https://github.com/finalburnneo/FBNeo/pull/1596

@dinkc64 It's okay.

PR #1604, opened 2023-11-29
Update d_spectrum.cpp

mobasweet:
Seraphima new version

---

Use proper version of this game which is current and available on developers website.


dinkc64:
But, is it better than the enhanced final version?

CorpseSlayer:
I ask in zosya.net if they will add the Seraphima_PT EFE and there answer was;

admin says:
no, as it was the same, but with some bugs. we don’t even have that version left.


So in short to answer your question:
> The version you added to FBNeo was basically remove from existence.
is no longer available, at least this is what i get when they say  "we don’t even have that version left"

So my subjection is to update to the latest final version located on
https://www.zosya.net that include the (RU) (PT) (EN) versions.

mobasweet:
> But, is it better than the enhanced final version?

Apparently the efv is an earlier build

mobasweet:
proper size for the rom added

PR #1623, opened 2024-01-01
Optimize the input problem of neogeo when using the keyboard

crashGG:
When the arcade stick input switches in two opposite directions, there is a "empty" in the middle, but when the emulator uses the keyboard to switch in two opposite directions, the switch is often instantaneous. For example, when switching left and right on an arcade joystick, the actual input is left->empty->right. That is to say, there will be an empty direction when switching to the opposite direction! But when using keyboard input, the actual input is left->right, or left->left+right->right

The problem I discovered is that some game developers deliberately added matching for the empty direction input when checking command inputs. For example, kof98 rugby charactor's 24A, 246A, 222B, 28A/C. For the last 28A/C , I found that the actual command is 2N8+A/C, that is, a total of 3 directions of detection: down, empty, up, and A/C! Once the empty direction is not detected, this command will fail, which is why it is difficult for most players to use the keyboard to do this trick!

This modification refers to the code of gsc2007 and pays tribute to him.

### DIFF
--- a/src/burn/drv/neogeo/neo_run.cpp
+++ b/src/burn/drv/neogeo/neo_run.cpp
@@ -129,6 +129,7 @@ UINT8 NeoJoy3[8]     = { 0, 0, 0, 0, 0, 0, 0, 0 };
 UINT8 NeoJoy4[8]     = { 0, 0, 0, 0, 0, 0, 0, 0 };
 UINT16 NeoAxis[2]	 = { 0, 0 };
 UINT8 NeoInput[32]   = { 0, };
+UINT8 NeoInput_previous[32]   = { 0, };	// GSC2007
 UINT8 NeoDiag[2]	 = { 0, 0 };
 UINT8 NeoDebugDip[2] = { 0, 0 };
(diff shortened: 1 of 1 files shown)


crashGG:
The game's "empty direction" input detection mechanism will affect all keyboard , button-type direction operations, such as hitbox, and some console joystick which opposite directions of cross keys can be pressed simultaneously at the same time.

dinkc64:
Interesting, can you tell where did this code from GSC2007 originally come from?

best regards,
- dink

crashGG:
> Interesting, can you tell where did this code from GSC2007 originally come from?
> 
> best regards,
> 
> * dink
His code comes from here:
https://www.ppxclub.com/forum.php?mod=viewthread&tid=671644
a Chinese emulator forum. The code was written in 2016 and was written for Fba and mame. I did a simple port to FBneo, and changed the comments to English.
GSC2007 explicitly expressed the hope that someone would submit this code to the relevant emulator mainline, but over the years this code seems to have been forgotten.

dinkc64:
Very nice, let's try it! :)

PR #1624, opened 2024-01-02
Optimize the input  of pgm when using the keyboard

crashGG:
the same as last commit
https://github.com/finalburnneo/FBNeo/commit/b89b79b75896ccccc500a64bafcd350b74487156

### DIFF
--- a/src/burn/drv/pgm/pgm_run.cpp
+++ b/src/burn/drv/pgm/pgm_run.cpp
@@ -11,6 +11,9 @@ UINT8 PgmJoy4[8] = {0,0,0,0,0,0,0,0};
 UINT8 PgmBtn1[8] = {0,0,0,0,0,0,0,0};
 UINT8 PgmBtn2[8] = {0,0,0,0,0,0,0,0};
 UINT8 PgmInput[9] = {0,0,0,0,0,0,0,0,0};
+UINT8 PgmInput_previous[9] = {0,0,0,0,0,0,0,0, 0};			//GSC2007  ADD
+UINT8 PgmInput_previous_lr[9] = {0,0,0,0,0,0,0,0, 0};		//GSC2007  ADD
+UINT8 PgmInput_previous_ud[9] = {0,0,0,0,0,0,0,0, 0};		//GSC2007  ADD
 UINT8 PgmReset = 0;
 static HoldCoin<4> hold_coin;
 
@@ -1008,6 +1011,37 @@ INT32 pgmExit()
 	return 0;
 }
 
+inline static void pgmClearOpposites(UINT8* nJoystickInputs,UINT8* nJoystickInputs_previous,UINT8* nJoystickInputs_previous_lr,UINT8* nJoystickInputs_previous_ud)	// GSC2007  add;
+{
+		if((*nJoystickInputs & 0x18) == 0x18)
+		{
+			 *nJoystickInputs &= *nJoystickInputs ^ *nJoystickInputs_previous_lr;//When left and right are pressed simultaneously, remove the interference items in *nJoystickInputs_previous_lr to obtain the actual required corrective input
+		} 
+		else if(*nJoystickInputs & 0x18)
+		{
+			*nJoystickInputs_previous_lr = *nJoystickInputs & 0x18;        //Only record the input when left and right are NOT pressed simultaneously
+		}
+		if((*nJoystickInputs & 0x06) == 0x06)
+		{
+			*nJoystickInputs &= *nJoystickInputs ^ *nJoystickInputs_previous_ud;        //Same as above
+		}
+		else if(*nJoystickInputs & 0x06)
+		{
+			*nJoystickInputs_previous_ud = *nJoystickInputs & 0x06;        //Same as above
+		}
+		
+//In extreme cases, opposite inputs are mutually exclusive
+		if ((*nJoystickInputs & 0x06) == 0x06)		
+			*nJoystickInputs &= ~0x06;
+		if ((*nJoystickInputs & 0x18) == 0x18)
+			*nJoystickInputs &= ~0x18;        
+
+		if (((*nJoystickInputs | *nJoystickInputs_previous) & 0x18) == 0x18)
+			*nJoystickInputs &= ~0x18;        //When switching left and right, the first frame is mutually exclusive
+		if (((*nJoystickInputs | *nJoystickInputs_previous) & 0x06) == 0x06)
+			*nJoystickInputs &= ~0x06;        //When switching down and up, the first frame is mutually exclusive
+}
+
 static void pgm_sprite_buffer()
 {
 	if (pgm_video_control & 0x0001) // verified
@@ -1039,6 +1073,9 @@ INT32 pgmFrame()
 
 	// compile inputs
 	{
+		for (INT32 i = 0; i < 4; i++) {
+			PgmInput_previous[i] = PgmInput[i];// GSC2007  add;
+		}
 		memset (PgmInput, 0, 6); // 6 is correct! Regions are stored in 7!
 
 		for (INT32 i = 0; i < sizeof(PgmJoy1); i++) {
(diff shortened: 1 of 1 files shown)


crashGG:
Since the input part of fbneo is separate for each driver, it will be very convenient if there is a unified module like mame.
I'm having trouble porting this code to the cps series driver, if anyone can help me it would be greatly appreciated.

barbudreadmon:
I have mixed feelings about crediting gsc2007 at each line endings in our source code, firstly because we don't do it for other contributors, secondly because iirc this guy used to break our license with some closed-source fork (and maybe still does ?), and even complained when we added support for some of his romhacks, saying they are his intellectual property (they are not, they belong to SNK).

crashGG:
You can modify the comments. GSC2007 is a selfless coder with a spirit of sharing, and he doesn't mind.

barbudreadmon:
>GSC2007 is a selfless coder with a spirit of sharing

He certainly didn't look like that when he requested we remove his hacks so that they'd be exclusives to his license-breaking fork...

barbudreadmon:
>You can modify the comments

It'd prefer not having those line endings in the first place, and simply having "[GSC2007]" at the end of the commit description, as it is usually done when porting code from another dev.

barbudreadmon:
@dinkc64 Maybe something akin to the "hold coin" functions could be done to standardize this ? Also i noticed this commit and the previous one probably don't work properly with runahead.

barbudreadmon:
I made an attempt at implementing those standardized functions at https://github.com/finalburnneo/FBNeo/commit/ef28bf02816dc2923b52afa0b426ed9cef9241db, i'm not entirely sure how to test it though.

dinkc64:
barbudreadmon, that's really nice to put it in joyprocess like that.  Though, I have no way of testing it, because I don't play fighting games.

crashGG:
> I made an attempt at implementing those standardized functions at [ef28bf0](https://github.com/finalburnneo/FBNeo/commit/ef28bf02816dc2923b52afa0b426ed9cef9241db) (it's in a separate branch, for testing), i'm not entirely sure how to test it though, and i think there is room for improvement whether it works or not.

This code works fine on neogeo driver(testing with kof97+2 frames runahead) using Retroarch + fbneo core, but I can't easily test how well it works on pgm driver. Because there are only two fighting games on the pgm driver that are not very familiar. Could you port it to the cps1 2 driver in the separate branch for testing? There are tons of fighting games on these two drives.They are also the games that players have been most familiar with for many years, and any improvement in operation will be immediately noticeable to us.

barbudreadmon:
@crashGG I'm extremely confused about the purpose of this code.

My understanding was that it would insert a frame with neither left nor right pressed when quickly switching from left to right. However i tried printing values with your changes from https://github.com/finalburnneo/FBNeo/pull/1623, and i can tell it doesn't do that : if i press left then right in kof94, `NeoInput[ 0]`'s value will go from 4 to 8 without going back to 0 in between. As a matter of fact, my standardized functions which i based on that PR are acting exactly the same.

Am i misunderstanding something about the purpose of this code ?

barbudreadmon:
To better explain, the original code (before https://github.com/finalburnneo/FBNeo/pull/1623) was cancelling both directions when opposite directions were pressed, meaning you'd actually have "left => nothing (neither left nor right pressed) => right" when pressing left then right on a keyboard, which seems to be what you tried to achieve with this PR. However your PR (and my code) is doing something entirely different : it replaces the direction pressed as soon as the opposite direction is pressed, meaning if you press right while still holding left, right will be pressed and left will be ignored.

crashGG:
> To better explain, the original code (before #1623) was cancelling both directions when opposite directions were pressed, meaning you'd actually have "left => nothing (neither left nor right pressed) => right" when pressing left then right on a keyboard, which seems to be what you tried to achieve with this PR (or maybe i simply misunderstood your explanation). However your PR (and my code) is doing something entirely different : it replaces the direction pressed as soon as the opposite direction is pressed, meaning if you press right while still holding left, right will be pressed and left will be ignored.

There is a way to test the effectiveness of this code. For a character using charging skills, such as Leona, first press the left button to charge, then press the right button without releasing the charge. At this time, the move can be completed. It proves that this code make an "empty" direction.Without this code, such a move operation cannot be completed.

barbudreadmon:
>There is a way to test the effectiveness of this code.

I'm not talking about effectiveness, i'm talking about what this code actually does. 
You can confirm yourself by adding
 ```
printf("%d: %x\n", nCurrentFrame, NeoInput[ 0]);
```
at the end of `NeoStandardInputs`, before and after your PR.

crashGG:
The author explained in the [post ](https://www.ppxclub.com/forum.php?mod=viewthread&tid=671644)that the mechanism of the code is block the first frame of the instantaneous switching

this code for two purposes:
1.Remove unnecessary interference input from keys;
2.Correct the missing and gaps in the semicircle input.

At present, most modified  fba/mame in the Chinese community have added this code. After being tested by millions of Chinese players, it has indeed improved the moves of keyboard players.

barbudreadmon:
>The author explained in the [post ](https://www.ppxclub.com/forum.php?mod=viewthread&tid=671644)that the mechanism of the code is block the first frame of the instantaneous switching

I don't speak chinese, and again that's not what this code does, you can easily confirm it yourself using the method i explained above.

I'm not asking if it improves moves or not, i'm asking if it's normal that it does the exact opposite of what you attempted to explain.

crashGG:
As I said above, I am not the author of this code and my understanding of it is limited. What I can do is bring the results of modification of the Chinese community to fbneo as the author hopes. I hope fbneo will be better, that's all. Maybe another fbneo contributor, @taoenwen, can explain this code better. He is also a member of the Chinese emu community.

barbudreadmon:
Ok, i'll try asking taoenwen on discord

crashGG:
Let’s be clear first: Arcade machines are designed for joystick players. Whether it’s a keyboard or a controller, when inputting moves, we are mostly simulating joystick input!
Keyboard users generally have this characteristic: they prefer multi-directional input and don’t pay much attention to redundant keys!
For example, what we call the big semicircle move: 246 (←↓→)
If A, S, and D are used to represent the direction, the most accurate input should actually be: A-AS-S-SD-D, which is the five directions, which is 41236 (←↙↓↘→)
It's easy for a keyboard user to go: A-AS-ASD-SD-D. What does this lead to? Depends on the situation
kwaks will be processed as ←↙↘→
mame and fba will be processed into ←↙↓↘→
So when using kawaks, there is a lack of arrow keys ↓! For games with fuzzy input like 97, the lack of ↓ has no effect on the game's moves. For strict moves like 02, the move will fail.
Therefore, the reason why it is difficult to make moves in the game: The essence is the habit of keyboard players! Subjectively we entered it, but in fact the input was wrong.
So optimizing moves, to put it bluntly, is actually correcting moves! Thereby improving the matching between player input and game instructions!
Therefore, move optimization is divided into rom level and simulator level.
But before that, we must first explain the reason why the kawaks simulator recognizes this way: kawaks does not process when the player presses the opposite direction keys at the same time! All inputs are given to the game. There is no simultaneous input of opposite keys on the joystick. The game is generally: input if left else right. There are left and right signals, and the left is given priority! Attached is the missing “↓” of ggxx:
QQ截图20130511215513 png thumb

Simulator level optimization methods:

As mentioned above, the reasons why it is difficult for kawaks to make moves. So how are mame and fba handled?
Both mame and fba block the opposite keys at the same time, so there are no these two actions.
Therefore, when you don’t know the most basic move situation of a game, press left and right at the same time, or press up and down at the same time, and you will understand by looking at its situation. The reason why TTX98UM has difficulty in making moves is that the loader does not handle pressing left and right at the same time. In the following situation, in the game, no matter you are on the 1p side or the 2p side, pressing left and right at the same time will always display left! But kof13 does not, because the kof13 program itself handles it and blocks these two inputs!
But this is the simplest and roughest optimization method. After experiencing the move optimization of Youju and ARC, I combined my own move habits, and the overall situation should be as follows: when pressing left, press right and 2 keys will appear. when press them together, which should be processed as right, that is, A-AD-AD-AD -D is corrected to A-D-D-D-D but when it goes from left, down, and right, it needs to be processed. AS-ASD is corrected to AS-S-SD, and then by analogy, it is expanded to 8 directions.

The horizontal direction is specifically processed into 4 parts:
1. For the first time, A to AD are processed into the opposite D;
2. Subsequent ADs will be processed into the same D;
3. The route from A to AS and then to ASD should be processed with the highest priority (here is the third item according to the explanation). If ASD is found for the first time, it should be processed as S, and then processed as SD;
4. Check whether A to AD are processed and then press S to see if s is valid. At this time, S and D should be valid, that is, SD, and should not be excessive S or D-SD-SD. Because of the general optimization of the simulator, there should be no problem with the game being misfired. for example,The character Moriya of The last blade has the move ←↙↓+A.
After all are processed, do the same from D to S, and then expand to the vertical direction!

I hope this article will be helpful to many players and the development of emulators.

                                                                                                                 - GSC2007

Finally, I found a detailed documentation of the code and translated it into English.

taoenwen:
I'm sorry it's taken me so long to get here. I've been bedridden for three more days with that damn flu.
I know I'm missing something, and it doesn't sound like things are that complicated, but joysticks and keyboards (or HITBOXes) do have some differences!

taoenwen:
The code submitted by crashGG solves the problem of making keyboard input (the uncertainty of the maximum number of arrow keys that can be pressed at once) reliable by determining whether the input of the previous frame conflicts with the input of the current frame and deciding whether to keep it or discard it.

taoenwen:
> I made an attempt at implementing those standardized functions at [ef28bf0](https://github.com/finalburnneo/FBNeo/commit/ef28bf02816dc2923b52afa0b426ed9cef9241db) (it's in a separate branch, for testing), i'm not entirely sure how to test it though, and i think there is room for improvement whether it works or not.

When you implement it upstream, it also facilitates the opportunity for other substrates to do it easily.

barbudreadmon:
>by determining whether the input of the previous frame conflicts with the input of the current frame and deciding whether to keep it or discard it.

Ok, that and the "detailed documentation" above seems to match my findings : this code is simply meant to give higher priority to latest pressed direction when opposite directions are being pressed at the same time. Now that i know it's the expected behavior, debugging will be easier.

barbudreadmon:
Ok, i'll close this PR and merge my work into the main branch, as far as i can tell from analyzing the input bits, it works properly, testers welcome though.

PR #1641, opened 2024-01-22
d_neogeo: samsho2pe v. 1.9

Igor-Arabe:
Removed bcd scaped command from, hanzo, kuroko and genan, moves scape from hanzo and kuroko It is now possible to use them in the middle of the game, new command in move list,  check!    Standardized Stun for attacks from above (jumping attack), stun medium for all chars! Fixed Kuroko talk before challenging to fight
Create new frame the char pick ups impulse before jumping, 2 frames for starting jump for all chars!

### DIFF
--- a/src/burn/drv/neogeo/d_neogeo.cpp
+++ b/src/burn/drv/neogeo/d_neogeo.cpp
@@ -20446,11 +20446,11 @@ struct BurnDriver BurnDrvsamsh2jq = {
 };
 
 
-// Samurai Shodown II Perfect Hack v. 1.8 - 2023-12-26
+// Samurai Shodown II Perfect Hack v. 1.9 - 2024-01-21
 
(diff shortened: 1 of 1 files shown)


dezraj:
Hello @Igor-Arabe !

The graphical glitch before Kuroko fight still occurs here, as you see below:
image

Also, the dialogue balloon is not filled with the Kuroko's words.

I'm playing the game in AES mode with Japanese language. I can provide a save state if you need.

Thank you so much for your work.

PS: though you see v.1.8 in the window title I'm actually using the latest version available in the FBNeo Archive.org set, which has the updated ROM files.

Igor-Arabe:
Thanks, but I only updated the CRCs. I contacted hack author and he answered that modifications is only for MVS mode. Any modification to fix that issue will be made only if not affects current hack (MVS mode).

PR #1686, opened 2024-02-21
d_cps1: fix input & dip in wofsgzb

taoenwen:


dinkc64:
I noticed that in the other bootleg, mentioned on the issue report, the flickery stuff didn't get fixed with rasterinit @ 12mhz, but, in fact the game just didn't like 12mhz at all.

taoenwen:
> I noticed that in the other bootleg, mentioned on the issue report, the flickery stuff didn't get fixed with rasterinit @ 12mhz, but, in fact the game just didn't like 12mhz at all.

We might as well observe how they work in the status quo for a while, wofsgzb revamping might be just what we need for this yet.

dinkc64:
sound good

PR #1733, opened 2024-03-23
v60: prevent unaligned read

barbudreadmon:
I have seen some weird reports about segas32 crashing, i *think* it might be due to the unaligned v60 reads i found in ubsan.

@dinkc64 this commit fixes the warnings and doesn't seem to introduce regressions, what do you think ?

### DIFF
--- a/src/cpu/v60/v60.cpp
+++ b/src/cpu/v60/v60.cpp
@@ -138,11 +138,11 @@ static UINT16 cpu_readop16(UINT32 a)
 	UINT8 *p = mem[2][a / page_size];
 
 	if (p) {
-		UINT16 *z = (UINT16*)(p + (a & page_mask));
+		UINT16 z = BURN_UNALIGNED_READ16(p + (a & page_mask));
 #ifdef LOG_MEM
-//		bprintf (0, _T("OP16: %6.6x %4.4x\n"), a, *z);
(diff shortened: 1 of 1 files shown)


dinkc64:
Hmm, can you show me the full reports?
Is it only happening on a system which doesn't allow unaligned reads for 16/32 bit accesses?

barbudreadmon:
There were 2 people on discord who said orunners is crashing on their android device, then another one on reddit who said slipstrm crashes (it turns out it hangs after asking more details but he has yet to tell me his OS). 
It reminded me of the issue some android users used to have with mslug5, svc and kof2003.

dinkc64:
Well, make z a const and have them try it? :)

dinkc64:
I mean, you don't have to const it, just merge it and see how it works out?

barbudreadmon:
After investigation, it turns out it's probably something else : 
- that last user was on windows 10, i think it might be something about his settings/drivers preventing resolution resize
- one of the 2 android users was experiencing those crashes only with netplay, so maybe the recent updates broke netplay backward compatibility
- not entirely sure about the other android user, but he said every games crashes including sf2, si i doubt it is actually something FBNeo-related

I'll close this.

dinkc64:
@barbudreadmon please see forum pm

Issue #1726, opened 2024-03-20
Are these hacks really good?


SumavisionQ5:
I found that recently some HBMame hacks have been inclueded into FBNeo (see: https://github.com/finalburnneo/FBNeo/issues/1715), and honestly speaking, those hacks are not fantastic. 
They are poorly hacked, and some of them are really boring. 
If anyone want to play those hacks, he can run them with IPS on Windows platform or run them by putting the patched version of the romset into SYSTEM_DIRECTORY/fbneo/patched with RetroArch on other platforms.
I've no intention of offending anyone, but I think FBNeo should be more strict with hacks.
Even now, HBMAME stops to include cheap hacks. 
FBNeo has many ways to run hacks, so we should not expand game library too prematurely.

taoenwen:
Thank you for your suggestions, good intentions don't offend anyone.

In this regard, we might as well listen to more voices as well, and if there is a consensus, we will certainly take it on board.

dinkc64:
well, not everyone has same taste, also NeilDark is a respected user
best regards,
- dink

Neildark:
By allusions, and also to explain my criterion, when I requested that these hacks be included, I would like to comment on some things, and excuse me for my poor English: in the taoenwen repository, https://github.com/taoenwen/FBNeo_IPS, which could be considered the official one for Final Burn NEO, there is no IPS for the games I added, there were no hacks for World Heroes 2, Word Heroes 2 Jet, Fatal Fury 2, Fatal Fury Special, Fatal Fury 3, Marvel Super Heroes Vs. Street Fighter, X-Men or Hyper Street Fighter II: The Anniversary Edition.

For fans of fighting games, as is my case, the Easy Moves trick, which is what I have chosen, is very necessary to do TAS and tests for bugs and curiosities in games that are then uploaded to YouTube, and I think it is very interesting especially in old games, and also to introduce the emulator. I'm aware that there are MAME IPS that could be used for this very thing, but not everyone has the time to find and manage them, and they don't always work.

As for Hyper Street Fighter II: The Anniversary Edition (040202U, 30% Blood) and fixed difficulty) (hack 2024), it is chosen because it is a very complicated game for new players and those two tricks help a lot to enjoy the game, the First, it reduces the damage of the hits and second, the difficulty, which, by the way, is a huge mistake in the original game because changing it has no effect.

[...]

PR #1734, opened 2024-03-24
add hack hsf2jpp/hsf2app

alhumbra:
Although a US version of hsf2pp already exists
Japanese (hsf2jpp) and Asian (hsf2app) versions have been added.

The hacker's name, Yumeji, and the date have been added to the ROM information
A script that can generate Roms in HTML can be downloaded from (Japanese website)
https://ux.getuploader.com/yumeji/download/21

screen shot
hsf2jpp

### DIFF
--- a/src/burn/drv/capcom/d_cps2.cpp
+++ b/src/burn/drv/capcom/d_cps2.cpp
@@ -12690,6 +12690,58 @@ static struct BurnRomInfo Hsf2ppRomDesc[] = {
 STD_ROM_PICK(Hsf2pp)
 STD_ROM_FN(Hsf2pp)
 
+static struct BurnRomInfo Hsf2jppRomDesc[] = {
+	{ "hsf2jpp.03",    0x080000, 0xb5952336, CPS2_PRG_68K | BRF_ESS | BRF_PRG },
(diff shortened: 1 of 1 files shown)


taoenwen:
https://github.com/finalburnneo/FBNeo/issues/1726

@alhumbra Is it necessary to add hacks that are only regionally different?

alhumbra:
Of course.
English is not the only language.

taoenwen:
> Of course. English is not the only language.

FBNeo offers a variety of possibilities, such as:
- ips
- romdata
All of the above methods can have the same effect.

barbudreadmon:
Adding 3 times the same hack because of language, for a game where the texts don't matter at all, feels a bit excessive.

dinkc64:
I think its ok, well, the only concern I have - why is the picture on the left side black? Was the patching successful?

alhumbra:
@dinkc64 Thank you for allowing the merge.

The patch matches the author's CRC, so it is probably not a problem.

Those captured from FBNeo were normal.
hsf2jpp-03-24-230315

Even the non-hacked version has black lines on both sides.
test000

I think it's the environment, GPU or blitter some other issue.

PR #1788, opened 2024-05-21
Update SFZ3Mix to .3

lolsuite:


taoenwen:
I suggest that you upgrade the version representation of that game along with it, do you think you can help?
Thanks！

taoenwen:
thanks

PR #1792, opened 2024-05-30
win32: Add input macros for System Next State and System Previous State

handledexception:
### Description
This PR adds input macros called "System Next State" and "System Previous State" to the Win32 build of FBNeo. These macros can then be bound to inputs, in the same manner as the existing "System Save State" and "System Load State" macros. The System Next State macro triggers the F11 hotkey, which selects the next save state slot, while the System Load State macro triggers the F8 hotkey, selecting the previous save state slot.

### Motivation
I got tired of having to take my hands off of my gamepad to press F8 and F11 to switch between save state slots. I wanted to be able to bind these shortcuts to buttons on my gamepad.

### How has this been tested?
I've tested this by building the Win32 build, mapping the new macros to gamepad buttons, and verifying that they perform the expected actions while running a game. Each macro is only trigged a single time for each button press, which is how the existing System Save State and System Load State macros behave.

### DIFF
--- a/src/burner/burner.h
+++ b/src/burner/burner.h
@@ -149,6 +149,8 @@ extern UINT8 macroSystemFFWD;
 extern UINT8 macroSystemFrame;
 extern UINT8 macroSystemSaveState;
 extern UINT8 macroSystemLoadState;
+extern UINT8 macroSystemNextState;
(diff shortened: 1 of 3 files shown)


dinkc64:
A nice addition, thanks! :)

PR #1822, opened 2024-06-29
Some changes in nes/pre90s

Igor-Arabe:
nes
d_nes - ghostsngoblins / ghostsngoblinsh: removed "NI 1.3" from header.
"NI 1.3 is the watermark of an old tool called NES Image converter." (by noone).

pre90s
d_pacman - mspacmane2: correct size for 'mpe6.6m' (checked in fbneo and rom manager).
d_tsamurai - alphaxz: changed "m660" to NULL, coz alphaxz is a parent ROM.


taoenwen:
thanks!

PR #1874, opened 2024-09-18
Enabled language switching in Tokio.

alhumbra:
The dip-switch has English and Japanese, but it did not work.
Enabled language switching in Tokio.

Not work.
dipsw_old

selectable.
dipsw

Switch to the Japanese version using the dip switch to check operation.
スクランブルフォーメーション (海賊版)


barbudreadmon:
Thanks !

PR #1914, opened 2024-11-17
Ignore SwapBytes.c (not used and has build error in SDL fork)

cjom:
When compiling SwapBytes.c (to be precise, cross compiling for armv7 was what I tested), fails with multiple errors like these:

`
src/dep/libs/lib7z/SwapBytes.c:598:47: error: expected ')' before ':' token
   #define SWAP2_32_VAR(v)  asm ("rev16 %0,%0" : "+r" (v)); // for clang/gcc
`

As pointed by @barbudreadmon , SwapBytes.c is not needed ( https://github.com/libretro/FBNeo/blob/master/src/burner/libretro/Makefile.common#L299-L314 ), so it should be safe to ignore.


barbudreadmon:
The lib7z library seems to have several platform compatibility issues, maybe this should also be reported there.

PR #1924, opened 2024-11-28
d_cps2: fix ddsomjr1, ddsomjr2 romSets

Igor-Arabe:
I found an inconsistency in japanese ddsom roms.
after checking cps2shock's site
current ddsomjr1 should be ddsomjr2 (no changes in labels). current ddsomjr2 should be ddsomjr1 (labels are: dd2j.03e /dd2j.04e / dd2.05e / dd2.06e).

This occurs in mame too.
\\\

image


dinkc64:
thanks for the romset fixes!

PR #1926, opened 2024-12-03
Add SDL2 set window scale command line option

grolliffe:
In the SDL2 front end, the window is scaled to * 2 the game screen resolution. Whilst this works find for raster games on normal DPI monitors it doesn't work so well for :-

- raster games on HiDPI displays where the window appears too small.

- vector games where the window appears too big.

This change allows the user to specify a window scale at the command line. It defaults to * 2 for backward compatibility with the old version.

Full screen and window resizing via dragging is not affected by this change.


barbudreadmon:
Thanks :)

PR #1928, opened 2024-12-17
Fix crash switching vector game resolution with SDL2

grolliffe:


barbudreadmon:
Hmmm, since other ports handle this without that change in `src/burn/vector.cpp`, i have a feeling this is not the correct way to fix this for thee sdl2 port.
At the very, in its current state, i'd expect this PR to require changes in other ports.

dinkc64:
Try this:
set pBurnDraw to NULL in Reinitialise() and check pBurnDraw != NULL in vector.cpp
This way pVidImage doesn't have to be extern'd into vector.cpp

best regards,
- dink

barbudreadmon:
I just checked and `draw_vector` is not even supposed to be called when resolution changed in a vector game : 
https://github.com/finalburnneo/FBNeo/blob/261417dcd1b8a3dbda8759fe26b2ef856db2104b/src/burn/drv/pre90s/d_bzone.cpp#L878-L882

Is it possibly a crash happening only with a specific vector game ? I only checked bzone tbh.

If not, there might be something very wrong happening in the sdl port.

grolliffe:
@dinkc64 : I used pVidImage because it's used in both platform and non-platform specific code to indicate if the buffer is valid. I wasn't sure if pBurnDraw would always be set to the same value (it does get set to things other than pVidImage). Switching to pBurnDraw and moving the test to the beginning of the function (my doh! for missing that optimisation) works.

@barbudreadmon : That caused me a lot of problems trying to figure out what is going on. I even tried mimicking what the WIN32 version does (post a message to the main event loop and do the reinit there) but that didn't work. I suspect something odd is happening related to the SDL event queue. Eventually I went for a belt and braces solution that does at least work and shouldn't affect other builds (if the buffer wasn't valid for another build, it would crash too so checking and not drawing if it isn't shouldn't cause problems). I should point out that I only run Linux, so I can't test how changes might effect builds for other OSs.

I've tried it with several vector games and the crash happens with all of them.

barbudreadmon:
>shouldn't affect other builds

Well, `pVidImage` is a variable purely meant for port logic, and doesn't necessarily exists in other ports. As a matter of fact, in its current state, the libretro port would break from your change.

barbudreadmon:
In the next few days, i'll try doing some debug on the sdl port in search of a better solution

grolliffe:
Thanks. This change was intended to fix an annoyance before implementing some other changes (support more resolutions, performance improvements, better bullets for Asteroids). I'll concentrate on them instead.

dinkc64:
@grolliffe neat, looking forward to your changes! :)

barbudreadmon:
>support more resolutions

As far as this is concerned, it has been on my mind for a long time and the best solution would probably be to have a width/height set by the port instead of the currently hardcoded 1920x1080. Those values could be set by the port through some new setting or by using the window size directly.

barbudreadmon:
Any tip on how i can reproduce that crash ? It doesn't seem to happen here. Are there specific settings to trigger this ?

FWIW, i'm using a desktop computer with some nvidia card and their proprietary drivers.

barbudreadmon:
I pushed the test fix, let me know if it improves things on your side.

barbudreadmon:
Ok, that's a weird one, this morning i did exactly that but i could never get the crash. 
5 hours and 1 reboot later, i'm trying again and getting the crash consistently.

`VidMem`,  `pVidImage` and `pBurnDraw` aren't exactly copies of each other, they are references to the same pointer, meaning if you modify one reference then the other references are also modified. They should be interchangeable in code, with `pBurnDraw` being the reference that you must use in the backend code.

I think i understand what's going on, in `vid_sdl2.cpp:Frame()` we are calling `Exit` then `Init`, it causes `VidMem` to be freed. Afterward i suppose `VidMem`, `pVidImage` and `pBurnDraw` stop refering to the same thing, which might be part of the problem.

2 different approachs could be done to fix this issue : 
- adding `pBurnDraw = NULL;` to `scrn.cpp:Reinitalise()` as you suggested, actually it's not even necessary to change anything in `vector.cpp:draw_vector()` if we do that.
- reallocating the pointer in `Frame` instead of calling `Exit` then `Init`.

The 1st approach is less work so that's what i'll do.

barbudreadmon:
New test fix pushed, does it work on your side ?

grolliffe:
> New test fix pushed, does it work on your side ?

Works here.

PR #1933, opened 2024-12-30
Add the MVS cart version of Captain Barrel (NeoByteForce)

arcadez2003:


arcadez2003:
Alright all i had was a romset minus any checksum info for the roms so im using the same boot 
values as Cyborg Force which is a total guess maybe i'll luck out and it'll work as it's by the same
company.

Hey it's been known to happen :)

PR #1941, opened 2025-01-24
Two of the TAPs don't "load"

lolsuite:
use "SD Card" Version of TAPs

### DIFF
--- a/src/burn/drv/spectrum/d_spectrum.cpp
+++ b/src/burn/drv/spectrum/d_spectrum.cpp
@@ -48550,7 +48550,7 @@ STD_ROM_FN(SpecSupalienbl)
 
 struct BurnDriver BurnSpecSupalienbl = {
 	"spec_supalienbl", NULL, "spec_spectrum", NULL, "2024",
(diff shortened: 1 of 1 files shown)


barbudreadmon:
So this is yet again the running gag where you send PRs to "update" some romsets to versions that don't work...

lolsuite:
Purchased rom came with 3 roms, tap,tzx and "sd card" tap.

since the tap don't load natively in FBNeo maybe the sd card will, even the readme says that.

dinkc64:
@lolsuite hi,
can you please link me the tzx's so I can figure out what's wrong with my new tzx loader code?
p.s. probably best to link them on neosource via pm or on rr :)

thanks,
 - dink

lolsuite:
it's weird, two were working, two are not. but check rr

lolsuite:
also note : the roms came with an emulator that require an activation key, but again two load (FBneo) and two don't.

dinkc64:
Can I get the .tzx's too?

dinkc64:
I fixed loading with the 2 games that wouldn't load: they use a custom loader, so they need to use the SlowTAP init which bypasses the dma loader.  I'm going to try to come up with a method to detect this stuff, so we don't have to use different inits for things like this...

best regards,
- dink

PR #1942, opened 2025-01-25
Update d_nes.cpp

lolsuite:
Galaxian is an overdump by default, [No-Intro](https://datomatic.no-intro.org/index.php?page=show_record&s=45&n=0811) , has a rom with split PRG (8kb/8kb) and then put together, this is smaller and still runs on emulators.

Completely Optional PR


dinkc64:
I can't get it to boot here, better stick with the old one

PR #1961, opened 2025-02-26
Arcade pre90s/tehkanwc

jjiglesias:
Hi! Here you have the driver "tehkanwc", ported from MAME's tehkanwc.cpp. It includes the following games: Tehkan World Cup, Gridiron Fight and Tee'd Off (and their clones).

The hardware has two analog trackball inputs. T've done my best to make them work figuring the values from MAME's and Advancemame's drivers out. I've tested them with a mouse and they seem to work properly. I've added fake digital inputs which can be used instead of the analog inputs.

In the driver scan function, I've used a pMin value taken directly from another driver. I don't know where this value comes from, so any help would be appreciated. And of course any edition too!

Tthanks to all FBN, MAME, and Advancemame developers: without you this driver wouldn't be possible!


barbudreadmon:
Thanks for your contribution ! However i can't merge this PR in its current state. 
Please send a PR containing only the new driver and the changes to makefile.burn_rules

barbudreadmon:
Some minor cleanups/changes i recommend for your driver : 
- replacing your multi-line memory init by `BurnAllocMemIndex` which handles it all.
- setting the refresh rate
- basing cycles-per-frame on the refresh rate (that one is arguable but sometimes it seems to allow to keep the correct game speed even after forcing 60Hz).
- removing most of those Hiscore function calls, the only function needed within a driver is `HiscoreReset` within the reset function, everything else is handled globally. => that one is mandatory

Here is a modified version with those changes : [d_tehkanwc.cpp.zip](https://github.com/user-attachments/files/18983671/d_tehkanwc.cpp.zip)

PR #1963, opened 2025-02-26
Added driver pre90s/d_tehkanwc.cpp

jjiglesias:
Hi! Here you have the driver "tehkanwc", ported from MAME's tehkanwc.cpp. It includes the following games: Tehkan World Cup, Gridiron Fight and Tee'd Off (and their clones).

The hardware has two analog trackball inputs. T've done my best to make them work figuring the values from MAME's and Advancemame's drivers out. I've tested them with a mouse and they seem to work properly. I've added fake digital inputs which can be used instead of the analog inputs.

In the driver scan function, I've used a pMin value taken directly from another driver. I don't know where this value comes from, so any help would be appreciated. And of course any edition too!

Tthanks to all FBN, MAME, and Advancemame developers: without you this driver wouldn't be possible!


dinkc64:
Hi @jjiglesias,
Wow, congrats on making a fbneo driver! :)  This is really great!

As barbudreadmon recommends, please remove calls to HiscoreInit(), HiscoreExit and HiscoreScan(), only HiscoreReset() is needed, the rest is handled by the burn library.  Or if you'd like, we can just merge this and I'll do it?

best regards,
- dink

barbudreadmon:
I think pnMin was supposed to be a versioning system for savestates, maybe meant to be used in "if pnMin loaded from savestate is lower than X, don't scan that variable" logic, allowing to load older savestates.

Well, basically it's useless since we never used it and there is a better method implemented nowaday.

jjiglesias:
Hi again! I've got an issue with shooting in Tehkan World Cup. This game has two kinds of ball shots: if you push the fire button firmly the shot is low, parallel to the ground. But if you push the button softly the shot is high, parabolic, so to speak. With current implementation, which include all the corrections and it's attached to this comment, I cannot get high shots.

[d_tehkanwc.zip](https://github.com/user-attachments/files/18989904/d_tehkanwc.zip)

barbudreadmon:
Hmm, assuming it's not an emulation problem (and i don't think it is), i guess you could create a fake button and implement some kind of logic to trigger the original input for a set number of frames behind that button. 
I'm not sure using the hold_coin functions for that purpose would be convenient, but feel free to do so if you think otherwise.

dinkc64:
Hi @jjiglesias,
The kick distance/height is determined by how fast the player is running (or moves the trackball) when kicking

Since the mouse doesn't really have momentum like a trackball, it's going to be impossible to play with the mouse.
But, if you scale up the analog inputs a little bit, it plays beautifully with a gamepad's analog thumbstick (I'm using a ds3 controller)

try this:
	BurnTrackballFrame(0, track_p1[0]*6, track_p1[1]*6, 0x2, 0x3f);  // 0x3f taken from advancemame driver

explanation: 2, 0x3f are just minimum/maximum clamping values for the accumulator (BurnTrackballRead value) movement per frame., to get a faster or higher signal, you'll need to use *x on the analog inputs.  After a bit of trial and error, I found *6 to be perfect for gamepad/thumbstick play.

best regards,
- dink

jjiglesias:
Hi @dinkc64 !

What you say about kick distance/height is true, but the kick button lets you perform two kinds of shots, which can be long or short depending on the trackball movement, as you said.

I was searching for some evidence of this in Internet, and I've found this web, [Tehkan World Cup - A Players' Reference Guide](http://www.kastellorizo.org/michael/reference.htm), which in "Section 1.1.1.2 : Passing/Shooting High and Low" says:

> The Kick Button can affect two modes of ball release: the high, flighted pass/shot and the low, driven pass/shot.

You can score a goal with a high kick from your half side of the playfield, actually.

I think we can consider movement done with your suggestion. In fact, I'm looking forward to come back home and try the game with an old thumbstick I have, who knows where, forgotten! Thank you very much!

dinkc64:
Also from this guide:
" to deliberately instruct a particular player to run in the direction of your choice. (This type of pass is not effective in the later rounds of a single player game, when the computer players are much faster than yours and will win most races to the ball in open play.) To specify a low, driven pass/shot, depress the Kick Button for at least a few tenths of a second. (To ensure the correct length of hold is achieved, you may elect simply to hold the Kick Button down until the pass/shot is released, as this will have the same effect.) To instigate a high, flighted pass/shot, a short tap of the Kick Button is required."

I did some playing and found the guide is right: the kick type entirely depends on how you press the button, and I don't think any extra code is needed to do this.  

I did some tests with this silly bit of code (for the debug log window):
	bprintf(0, _T("%x %S\n"), TWCInputPort0[5], TWCInputPort0[5] ? "<---" : "");
A short press (about 3 frames) makes the ball really fly high and a long one (10+ frames) makes a low flying kick

best regards,
- dink

barbudreadmon:
Hmmm, from the "coin input" issues i've seen over the years, i fear pressing a button for only 3 frames might be an issue for some users. It might really be a good idea to have a shortcut for the short press.

dinkc64:
Hey guys, I did more thorough testing regarding the kick button, here's the results:

button held for....
~~
11+ frames: "low, driven pass/shot"
10 and under: "high, flighted pass/shot"

I don't think 2 buttons is needed here, but, let's see what jjiglesias thinks :)

best regards,9
- dink

jjiglesias:
Hi!

It seems that it's a matter of getting the right value for the `VelocityMax` parameter:

BurnTrackballFrame(0, track_p1[0]*6, track_p1[1]*6, 0x2, **0x3f**);

I've noticed that when using the keyboard as analog input, I got a little elevation when "soft-pressing" the button. And zero elevation when using the keyboard as fake digital input. Fake digital inputs are supposed to get the maximun value out of the analog controls, so we are not getting this maximum with current 0x3f value.

I've tried several values and it's with 0x80 that I get the maximum 0x3f value. And I get a similar behaviour as in MAME with "fake" keyboard inputs: playing as P1 I can reach the goal from the bottom end of the central circle.

Would you please try

BurnTrackballFrame(0, track_p1[0]*6, track_p1[1]*6, 0x2, **0x80**);

with the thumbstick?

Thanks in advance.
Best regards!!

jjiglesias:
[d_tehkanwc.zip](https://github.com/user-attachments/files/19014987/d_tehkanwc.zip)

Please proceed with the PR if you see fit.

dinkc64:
Nice findings, I'll be able to test it some more later tonight! :)
fyi, if you want to just modify the digital value added to the accumulator, you could add it as the last parameter to BurnTrackballUDLR()

If it turns out 0x3f is the perfect maximal value for analog, but the digital value needs to be a bit faster, it would be useful in that case.  Just something to keep in mind :)

best regards,
- dink

dinkc64:
Hi @jjiglesias,
I did some playingtesting and tweaking values, etc, tonight and found these work the best:
	BurnTrackballFrame(0, track_p1[0]*6, track_p1[1]*6, 0x2, 0x3f);
	BurnTrackballUDLR(0, TWCFakeInputPort[3], TWCFakeInputPort[2], TWCFakeInputPort[1], TWCFakeInputPort[0], 0x3f);

I had to go back to 0x3f for BurnTrackballFrame()'s velocityMax parameter.  Using 0x80, a too fast movement could cause the game to go crazy.  (f. ex, set the Analog Sensitivity to 400% in "map game inputs" and try :) )  Specifying 0x3f for the last param. on BurnTrackballUDLR() will get the digital inputs just right.

Game is really fun and challenging for an early ('85) game!  Surprising how good the enemy intelligence is.  I'm really glad you brought this to fbneo! :)

best regards,
- dink

jjiglesias:
Here you have, @dinkc64 ! Thank you very much!
[d_tehkanwc.zip](https://github.com/user-attachments/files/19022833/d_tehkanwc.zip)

dinkc64:
Hi @jjiglesias,
I did some playtesting, and all seems well :)

Do you have anything against changing the digital controls to the standard up/down/left/right, instead of x/y inc/dec?

best regards,
- dink

jjiglesias:
Nothing at all! I don't know if this is the right thing, but would you please do it?

jjiglesias:
Hey @dinkc64 ! Standard control names done!

dinkc64:
Exceptional! :)

PR #1968, opened 2025-03-01
Update FamiDash to 1.2 + add "Side B"

lolsuite:


dinkc64:
Please fix your editor so it doesn't mess with the codepage of non-English languages, I merged it manually for now.

PR #1982, opened 2025-03-19
taitof3 updates

barbudreadmon:
@dinkc64 what do you think of those changes ?

### DIFF
--- a/src/burn/drv/taito/d_taitof3.cpp
+++ b/src/burn/drv/taito/d_taitof3.cpp
@@ -198,68 +198,65 @@ static struct BurnInputInfo KnInputList[] = {
 
 STDINPUTINFO(Kn)
 
-static struct BurnDIPInfo gseekerDIPList[]=
+static struct BurnDIPInfo F3GlobalDIPList[]=
 {
-	{0x2e, 0xff, 0xff, 0x01, NULL },
-
 	{0   , 0xfe, 0   , 2   , "Music Tempo (must restart!)" },
-	{0x2e, 0x01, 0x02, 0x00, "Normal / Fast" },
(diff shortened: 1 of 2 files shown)


dinkc64:
It looks good, but:
I remember that there was a game that crashed while processing sprites if the fps changed, but, maybe that's not an issue any more?  wish I could remember the game o.O

barbudreadmon:
Hmmm, an older version of the "do not touch" comment was mentioning freezes in dariusg & gekiridn, but i didn't get any when i played a few stages in those games.

barbudreadmon:
I'll try to do full runs of those games using invincibility cheats before merging

dinkc64:
My memory is a bit cloudy here, but:
I sorta remember that the freezes happened during/at bootup in a game (or 2?), I think sprites were linking back unto themselves, creating an endless loop.  & I think it was happening due to a certain pattern getting written to memory during the test while rendering.

barbudreadmon:
Ok, that issue definitely disappeared then.
I did a full gekiridn run and no issue, i'll skip the full dariusg run if the issue was supposed to be at boot.
Let's merge this :)

barbudreadmon:
note: I also updated cheat.dat at https://github.com/finalburnneo/FBNeo-cheats

dinkc64:
sounds good :)
Thanks for the heads-up re: cheat.dat !

barbudreadmon:
I don't think i'll find the time to look into this in the next 10 days or so, but i can give it a try afterward

dinkc64:
I've put it off since the accidental rm (months ago), there's no rush :)

dinkc64:
When you start, please let me know, incase I dive into anything.  I'll also do the same

PR #1996, opened 2025-03-31
fix: macOS build and Xcode files

frangarcj:
Make Xcode Build great again


dinkc64:
Looks good!
A small request/suggestion: 
In order to keep things working after another update of zlib,
add a file called update.txt to src/dep/libs/zlib
in the file, make a quick description of the changes to gzguts.h to keep the mac build working 

What do you think?

best regards,
- dink

PR #2015, opened 2025-04-20
Update d_megadrive.cpp

1eye4eye:
add cursedlegacy
add p47iimd


barbudreadmon:
It seems you did something wrong, probably with line endings

PR #2017, opened 2025-04-26
sf2prime updated CRC

lolsuite:
unsure about size attached picture for sizes.
crc image


dinkc64:
Get AnalogX PCalc to help with hex values, f. ex:
0x80000  = 524288 (512k)
0x800000 = 8megs

pcalc here: http://www.analogx.com/contents/download/programming/pcalc/freeware.htm

it's very handy and one of my most used tools other than my source code editor.

best regards,
- dink

PR #2023, opened 2025-05-08
Fix crash when no game selected

fjtrujy:
This avoids the crash when trying to open the menu if no game loaded
Screenshot 2025-05-08 at 17 00 17

I will try to fix the remaining options in a separate PR


taoenwen:
@fjtrujy Thanks!

PR #2040, opened 2025-05-17
sshot.cpp: replace fixed titles and support strings by the global variables that contains them

fprietog:
The user can choose his own folders instead of the default ones for titles and previews images. 
I've changed the default fixed strings used for these folders ("support/titles" and "support/previews") by the global variables that contains the current folders in use (szAppTitlesPath and szAppPreviewsPath).

### DIFF
--- a/src/burner/sshot.cpp
+++ b/src/burner/sshot.cpp
@@ -8,12 +8,8 @@
 
 #if defined(BUILD_SDL2) && !defined(SDL_WINDOWS)
 	TCHAR *SSHOT_DIRECTORY = NULL;
-	TCHAR *STSHOT_DIRECTORY = NULL;
-	TCHAR *SPSHOT_DIRECTORY = NULL;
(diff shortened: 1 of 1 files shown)


dinkc64:
Hi fprietog,
This is not a big deal, but a curiosity:
In your PR's, it always says "Co-authored-by: fprietog <fprietog@fpgrpi>".  Is your git client logged into the same account as your github account?

best regards,
- dink

fprietog:
The account is the same but when I tried to use the same email it gives me an error when on push saying that it will shown my real email and my actual security options forbids it.

So I used a ficticious one.

Captura desde 2025-05-18 02-47-58

If it's causing an issue I can check what security option I have that prevents it.

fprietog:
I've unchecked the option that prevents the use of the real email... We will see if it fixed the problem the next commit.

Thanks and best regards.

PR #2049, opened 2025-05-20
d_snes.cpp: Added the support to 5 new SNES/Super Famicom games

fprietog:
- Barbie - Super Model (USA)
- Barbie - Vacation Adventure (USA, Prototype)
- Hilda (HB)
- Incognity: A rover story (HB)
- Super Chase H.Q. (Australia)


taoenwen:
Thanks!

PR #2055, opened 2025-05-21
d_megadrive.cpp: bssf2gq v2.8

KlaviculaAG:


BelvaCompany:
Thanks. Is the updated No Pause subpatch that works for CPU matches in Tournament Mode used?

PR #2062, opened 2025-05-23
d_snes: chavezii typo

KlaviculaAG:
ROMset name should be in lowercase


fprietog:
Agggg! I didn't saw it :see_no_evil: 

Thanks for fixing!

PR #2070, opened 2025-05-25
d_snes.cpp: added the support to 10 new SNES/Super Famicom games

fprietog:
- BASS Masters Classic (USA)
- BASS Masters Classic - Pro Edition (USA)
- Captain Novolin (USA)
- Jimmy Houston's Bass Tournament U.S.A. (USA)
- Lawnmower Man, The (USA)
- Mario Is Missing! (USA)
- Mark Davis' The Fishing Master (USA)
- TNN Bass Tournament of Champions (USA)
- Where in the World Is Carmen Sandiego (USA)
- Where in Time Is Carmen Sandiego (USA)

### DIFF
--- a/src/burn/drv/snes/d_snes.cpp
+++ b/src/burn/drv/snes/d_snes.cpp
@@ -3121,6 +3121,44 @@ struct BurnDriver BurnDrvsnes_Barkleyj = {
 	512, 448, 4, 3
 };
 
+// BASS Masters Classic (USA)
+
+static struct BurnRomInfo snes_BassmastersclRomDesc[] = {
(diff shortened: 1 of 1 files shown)


fprietog:
> Are BASS Masters Classic (USA) and BASS Masters Classic - Pro Edition (USA) sure they are both standalone and not Clone?

Good question. 

This case is similar to "Wheel of Fortune" and "Wheel of Fortune - Deluxe Edition". Very similar games released in the same region with a year of difference. That's why I considered them not parent/clone related. 

Additionally:

- No-Intro doesn't consider them parent/clone related too.

- You have both "BASS Masters Classic" and "BASS Masters Classic Pro Edition" available in FBNeo for Megadrive and there aren't related there too.

Thanks and best regards.

PR #2086, opened 2025-05-31
Adj. in d_snes.cpp and d_megadrive.cpp

KlaviculaAG:
Adjustments in following games...
md_bssf2gq - update to v2.9A
snes_tornekodungte - update in addendum


taoenwen:
Thanks!

PR #2099, opened 2025-06-06
adj. in d_cps1.cpp, d_snes.cpp

KlaviculaAG:
d_cps1.cpp - dinosyn typo (cdj_syn.10f)
d_snes.cpp - add fireemblem776ti


taoenwen:
Thanks!

PR #2116, opened 2025-06-14
d_labyrunr.cpp: hook tilemaps

barbudreadmon:
@dinkc64 i'm not super confident about this one, even though it _seems_ to work properly, could you take a look ?

### DIFF
--- a/src/burn/drv/konami/d_labyrunr.cpp
+++ b/src/burn/drv/konami/d_labyrunr.cpp
@@ -6,6 +6,7 @@
 #include "burn_ym2203.h"
 #include "konamiic.h"
 #include "k007121.h"
+#include "rectangle.h"
 
 static UINT8 *AllMem;
 static UINT8 *MemEnd;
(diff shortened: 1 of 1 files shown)


dinkc64:
Sure, checking.  There's a few spots that really break if something is wrong - the ending, and the scrolling face/wall boss.  I'll check those first :)

dinkc64:
Unfortunately, this totally breaks the ending sequence.  Now I remember, the ending sequence is absolute insanity to deal with, I think there was like... 20 attempts to fix this over the years, and I finally got it a few years back.  Well, with that being said, let's see if it's something that can be fixed :)

Here's a state to the ending: 
[tricktrp slot 01.zip](https://github.com/user-attachments/files/20743130/tricktrp.slot.01.zip)

dinkc64:
This is going to be tricky...

barbudreadmon:
Hmm, how about we just don't fix what is not broken then ? 
I wanted to get those k007121 drivers the most up-to-date possible against mame but tbh they are already fine as they are.

dinkc64:
AGREED.  yea, my head is about to break :)

dinkc64:
Maybe close this for now?

PR #2119, opened 2025-06-15
d_snes.cpp - targa: missing info

KlaviculaAG:
targa is clone of rrangerr2

### DIFF
--- a/src/burn/drv/snes/d_snes.cpp
+++ b/src/burn/drv/snes/d_snes.cpp
@@ -1,5 +1,5 @@
 // FBNeo SNES driver / interface to LakeSnes
-// By Kev, dink, Gab75, fprietog
+// By Kev, dink, Gab75
 
 #include "tiles_generic.h"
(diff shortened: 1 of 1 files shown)


taoenwen:
On line 2, the driver author information has changed.

The previous change came from:

https://github.com/finalburnneo/FBNeo/pull/2115/commits/2f5f5a8138970b2a0cd41adb9ea9b55a0ade671b

Regarding the driver author change, it would be appropriate to have dink confirm this.

Personally, I think the current changes to this PR are correct.

@dinkc64 Sir.

fprietog:
That's obviously my fault as I viewed this concrete file d_snes.cpp as part of the driver, the supported games list part, and not the whole snes driver, so I just put myself s contributor after 500 games included. 

Sorry for the problem caused.

taoenwen:
Since we all have full judgment on this, there is no need for dink to arbitrate, thanks!

PR #2131, opened 2025-06-23
macos build fix

warmenhoven:


dinkc64:
Hi @warmenhoven 
Can you explain how those hashes work?  I wonder if someone can add new files to the build for macos from the linux or windows side?

barbudreadmon:
@warmenhoven Thanks for the PR !

@dinkc64 I got in touch with warmenhoven yesterday on discord (he is maintaining the libretro stuff on apple devices) and he explained to me that people usually don't do that manually...

I tried nonetheless after reading [this](https://behradkazemi.medium.com/add-a-file-to-pbxproj-in-programmatically-way-ff9661e6b0fc), but it failed and i reverted after 2 commits. 

I think i understand my mistake now, you need to generate 2 hashes per file, i only generated 1.
Basically in the first section (PBXBuildFile) you got "hash1 blahblah hash2"
2nd (PBXFileReference) and 3rd (the one with folder names) sections use hash2
Then the last section (PBXSourcesBuildPhase) use hash1 again.
The hashes seem to be 24 characters long hexadecimals, without any actual constraint (aside from being unique).

Now the real question is whether we want to re-enable this on the buildbot or not, i think this is very cumbersome for a port that is barely functional.

PR #2132, opened 2025-06-24
d_megadrive.cpp: add zpf

1eye4eye:


barbudreadmon:
Is it the demo or the full game ?
If that's the demo, it might be a good idea to mention it in the description & romset name, and also to add the BDF_DEMO flag.

I could be wrong but i think the game hasn't been released yet, so i assume that's the demo ?

1eye4eye:
this is full, i do not make demo shit
make it BDF_HORSHOOT

dinkc64:
thanks!

PR #2140, opened 2025-06-29
adj. in d_nes.cpp - FDS: add oprjumpout

KlaviculaAG:
Add Operation: Jump Out, english translation of Tobidase Daisakusen


KlaviculaAG:
crc patch`*`: 25cbbe75 -> caa25a7f

25cbbe75 Is CRC32 of fbneo ROM without header.
`*`Use this patch only if you didn't find ROM mentioned on translation site (caa25a7f).



[crc_patch.zip](https://github.com/user-attachments/files/20969692/crc_patch.zip)

PR #2150, opened 2025-07-15
adj. in d_nes.cpp, nes.cpp - ikariwarriorse

KlaviculaAG:
adj. in ikariwarriorse, add CRC to NesMode List (IS_PAL).


taoenwen:
thanks

PR #2165, opened 2025-07-28
d_snes: typo in legendofzeldae, legendofzeldatsc, legendofzeldati

Igor-Arabe:
removed invalid charachter (":") from fileName.

hack redux: name scheme as on parent (Legend of Zelda, The: A Link to the Past)


taoenwen:
thanks

PR #2175, opened 2025-08-06
House Keeping

EGudsson:
Update "Earthion" + "Creeping It Real"


dinkc64:
Looks like eye1eye already added earthion update.  Sorry about the conflict!

does the new rom of creeping it real display graphics fine when the guy dances?  What is the source of this rom?

best regards,
- dink

dinkc64:
btw, I already fixed creeping it real's mirroring in a previous merge to d_nes..
Simple edit: 0x40 -> 0x41 byte 7 (or 6 if zero based)

I'm confused here, is this another version of creeping it real?  There's a lot more stuff changed in the rom than the mirroring..


best regards,
- dink

PR #2194, opened 2025-08-23
  Add DAT generation support for 16 additional arcade systems

battousai90:
- Add CPS1, CPS2, CPS3, PGM, Cave, Taito, Toaplan, Psikyo system DATs
  - Add Konami, Irem, Data East, Sega Arcade, Midway, Galaxian system DATs

  - Add Pre-1990s and Post-1990s arcade categorization DATs
  - Expand from 17 to 33 total supported systems for DAT generation
  - Update burner.h with new DAT_*_ONLY constants (17-32)
  - Update dat.cpp with filtering logic and descriptions for all new systems
  - Update SDL main.cpp to generate all new DAT files via -dat command


barbudreadmon:
We won't add additional entries to generate dat files, there would never be an end to this.
[datutil](https://neo-source.com/index.php?topic=2487.msg35582#msg35582) is your friend, it does exactly what you want, and more.

PR #2216, opened 2025-09-13
Genesis rom update

EGudsson:
0.5b > 1.4


dinkc64:
next time use the alligator method :)

PR #2220, opened 2025-09-16
Segas32 fix text flip

mahoneyt944:


barbudreadmon:
Good catch ! Thanks :)

mahoneyt944:
Traced it back to around MAME 0.97u5, a 20+ year old typo lol.

PR #2231, opened 2025-09-23
Fix text resolution hack flip

mahoneyt944:
Have not tested this. I think it should be correct though. Please test before committing.


dinkc64:
Hi, your fix makes things look better, but the flipped offsets are a bit off.  I'll try to look into it later when I have some time

mahoneyt944:
Glad it's closer. I'm sure it's something minor. Unfortunately I don't have a build environment for FBNEO to test it out and see what's happening. Probably just need to account for the screen width or pixel shift from the right side with the resolution hack. Curious to see what you come up with.

dinkc64:
If you'd like, I can help you get everything needed to compile fbneo, I've kinda streamlined my methods over the years

mahoneyt944:
I would like to, however I code everything from my mobile phone so a compile environment isn't in the cards unfortunately. I appreciate it though.

PR #2235, opened 2025-09-25
d_xunit.cpp: increase cmos's default volume

barbudreadmon:
@dinkc64 i'm not too happy with the default volume for revx and was trying to find a better compromise. What do you think of this ? It'll only increase volume on a fresh nvram with the default setting.


dinkc64:
Well, this is decent, and looks good :)

PR #2261, opened 2025-10-08
d_megadrive.cpp: added the support to 'ColocoDX (HB)'

fprietog:
I don't know if it's working 100% perfect but it seem to works. Need testing.


fprietog:
md_colocodx-10-08-094750 
md_colocodx-10-08-095419

fprietog:
Music doesn't work well. Discarded.

PR #2269, opened 2025-10-19
d_neogeo.cpp: add maglordacfh

1eye4eye:


src/burn/drv/neogeo/d_neogeo.cpp
JacKc029735: Hi 1eye4eye,

Just added before i saw your pull  request: https://github.com/finalburnneo/FBNeo/commit/b81d69b489689dbb7f8d4b7b57f1cc9538bc796f

PR #2293, opened 2025-11-01
d_nes.cpp: added the support to "Splatter World (Japan, Prototype)"

fprietog:
Hi @dinkc64 ,

There is no need to add a new mapper for "Splatter World (Japan, Prototype)". It now works out of the box :ok_hand: 

Also added the support to "Fruity Pals' Revenge (HB, v1.2)" for NeoGeo Pocket Color, an unexpected but welcomed sequel to "Don't Die Mr. Robot".

Thanks and best regards.


dinkc64:
great news! :)

PR #2315, opened 2025-11-19
vector resolution wip

barbudreadmon:
This is the code i used on my 1440p screen to test the callback : 
```c
static INT32 screen_resize_applied = 0;
static INT32 screen_resize_callback(INT32 &width, INT32 &height)
{
	if (screen_resize_applied) return 0;
	int game_aspect_x, game_aspect_y;
	BurnDrvGetAspect(&game_aspect_x, &game_aspect_y);
	if (BurnDrvGetFlags() & BDF_ORIENTATION_VERTICAL) {
		width = 1440;
		height = width / game_aspect_y * game_aspect_x;
	} else {
		height = 1440;
		width = height / game_aspect_y * game_aspect_x;
	}
	screen_resize_applied = 1;
	return 1;
}
```

The callback pass width and height to the backend, while returning 1 if that resolution is expected to be different from previous frame, or 0 otherwise. I'd still need to make the 1440 dynamic, and set `screen_resize_applied` to 1 every time it changes.

And you declare the callback like this before initializing a driver : 
```c
BurnResizeCallback = screen_resize_callback;
```

If the callback is not declared, it just uses the default resolution from the driver.

@dinkc64 what do you think ?

### DIFF
--- a/src/burn/vector.cpp
+++ b/src/burn/vector.cpp
@@ -31,6 +31,8 @@ static float vector_intens      = 1.0;
 static INT32 vector_antialias   = 1;
 static INT32 vector_beam        = 0x0001f65e; // 16.16 beam width
 
+static INT32 vector_rescaled    = 0;
+
 #define CLAMP8(x) do { if (x > 0xff) x = 0xff; if (x < 0) x = 0; } while (0)
 
 static UINT8 gammaLUT[256];
@@ -77,6 +79,7 @@ void vector_set_scale(INT32 x, INT32 y)
 
 void vector_rescale(INT32 x, INT32 y)
 {
+	vector_rescaled = 1;
 	GenericTilesSetClipRaw(0, x, 0, y);
 	BurnTransferSetDimensions(x, y);
 	BurnDrvSetVisibleSize(x, y);
@@ -280,6 +283,11 @@ void vector_set_pix_cb(UINT32 (*cb)(INT32 x, INT32 y, UINT32 color))
 
 void draw_vector(UINT32 *palette)
 {
+	if (vector_rescaled) {
+		// don't draw this time around
+		vector_rescaled = 0;
+		return;
+	}
 	struct vector_line *ptr = &vector_table[0];
 
 	INT32 prev_x = 0, prev_y = 0;


dinkc64:
I was thinking of something more along the lines of this for the callback:

When the resolution of the screen/window/fullscreen changes:
We could call this function below from burner,
or even intf/video somewhere, if its more applicable:

void BurnScreenResized(int new_screen_width, int new_screen_height)
{
    if (BurnResizeCallback) BurnResizeCallback(new_screen_width, new_screen_height);
}

..and then vector would be rescaled already before the next draw.

best regards,
- dink

barbudreadmon:
Hmmm, then we'd need BurnResizeCallback to directly call vector_rescale ?

dinkc64:
Yes :)

barbudreadmon:
Hmmm, i'm missing the point of the callback then ? We call `BurnScreenResized`, which could be directly calling `vector_rescale` without going through a callback ?

dinkc64:
The way you're doing it: why even put the callback outside of vector.cpp?  I'm trying to understand
the sense in that.  

the idea/logic in my version:
in vector_init(), we'll set the callback to point to the function to compute the new res & call vector_rescale()
when the screen is created, resized, go to fullscreen,etc, BurnScreenResized() will be called with the
new size of the screen/window.
In this function, if the callback is set, it'll run it and tell the vector engine to resize

BurnScreenResize() only gets called when a resize event occurs (from Burner side, or intf/video), instead of
checking every frame.  It just makes more sense this way

best regards,
- dink

barbudreadmon:
But with your version, since the port code is triggering the resize while sending width/height at the same time, the function computing the new res & calling vector_rescale doesn't need to be platform-dependent, right ? 
I like your version better, but i'm still getting confused by the need for a callback function.

dinkc64:
I don't know what else to say or do at this point, other than merge what you got and I'll set it up and see how it works.

best regards,
- dink

barbudreadmon:
Aside from the callback thingy i didn't figure out, i like your idea better, it makes more sense than checking resolution at every frames like my PR does. 
I'll just close this PR and wait for yours.

dinkc64:
I want to merge this because you have half the work already done :)

grant2258:
You cant check the viewport resolution in retroarch anyway. At least since i last used it you could check only the framebuffer size. It would be ideal to scale to the viewport though but it is what it is I guess.

barbudreadmon:
Yes you can't, that's why several cores (including yours ?) dealt with this through a core option with pre-defined resolutions.

barbudreadmon:
>I want to merge this because you have half the work already done :)

It might be better to merge this once everything is ready since it removes the current hires dipswitch ?

grant2258:
yea we set fixed resolutions i did test 4k and it ran fine on a pi 5 but honestly unless there is demand for it, ill hold on ra providing the current viewport its using.

dinkc64:
> You cant check the viewport resolution in retroarch anyway. At least since i last used it you could check only the framebuffer size. It would be ideal to scale to the viewport though but it is what it is I guess.

Simply put, that limitation is re-f*cking-tarted. :(

dinkc64:
Well, there's little point to doing any of this here, other than adding another resolution to the dips

barbudreadmon:
There is a point for standalone and that's all that matter :)
For libretro it'll end up being not so different from the dipswitch, but still an improvement because adding more resolutions will be easier.

barbudreadmon:
@dinkc64 i think i finally understood the point of the callback, i'll modify my code later today and reopen the PR

barbudreadmon:
I think it's ok now ? 
note: skipping frame when resizing seemed required on my side, but that might be due to some logic specific to the libretro port

dinkc64:
Looks good, I'll do some testing & stuff with it after its merged (I don't like to mess around with my local git repo, especially branches and stuff, caused me a lot of pain in the past..)

barbudreadmon:
Ok, i'll merge it now then

PR #2332, opened 2025-12-03
d_neogeo zetsuxeving typo

Igor-Arabe:
correct year (2025)

image


dinkc64:
The last version was 2035 on titlescreen :)

PR #2421, opened 2026-01-09
d_snes.cpp: Super Ghouls'n Ghosts...

fprietog:
Added "Super Ghouls'n Ghosts (USA, Prototype)"
Updated "Super Ghouls'n Ghosts Restoration (Hack)"


dinkc64:
epic adds, thanks!!

PR #2499, opened 2026-04-10
Update pgm2： A new card will be inserted by default if no card is inserted.

crashGG:
A temp new card will be inserted by default if no card is inserted.

This allows the use of special characters that can only be used when a card is inserted. It does not affect normal card insertion; it only takes effect when no card is inserted.

### DIFF
--- a/src/burn/drv/pgm2/pgm2_run.cpp
+++ b/src/burn/drv/pgm2/pgm2_run.cpp
@@ -841,7 +841,13 @@ static void pgm2McuCommand(bool isCommand)
             break;
 
             case 0xc0: // insert card / check card presence
-            case 0xc1: // check ready/busy
+                if (!pgm2CardPresent(arg1 & 3)) {
+                    status = 0x00f70000;
+                }
+                Pgm2McuResult0 = cmd;
+			break;
+
+			case 0xc1: // check ready/busy
                 if (!pgm2CardPresent(arg1)) {
                     status = 0x00f40000;
                 }


dinkc64:
I don't know much about the memory card situation, so bear with me...
We have to think about the side-effects this might cause (if any)
Questions:
Is this a good idea?
Can this temp card be saved / will it be saved by default?
What happens if someone tries to insert or remove a card with this temp one
already active?  ..etc.?
Is it synchronized to the menu options?

@hogenwang What do you think about this?

best regards,
- dink

crashGG:
If you don’t know much memory cards, you’re an old‑school hardcore player just like me.
Early arcades were true arenas of skill. One coin, fresh start, equal footing for everyone.
The high-score table was our wall of honor.
 
Once memory cards arrived, everything changed.
Levels, gear, endless progression… arcades went soft, turned home‑console, lost their purity.
That raw, instant thrill faded away.
What was once pure skill became just another grind.

dinkc64:
crashGG,
I'm entirely with you, in that regards.  My time was early to mid-80s,
Spy Hunter, Qbert, Pac-Man, Defender, Tempest, etc, etc :)

best regards,
- dink

crashGG:
dink,
I got into arcade games a little later than you, in the mid-to-late 1980s. My first arcade game, as far as I remember, was  Gemini.Fortunately, I experienced the entire golden age of arcade. This experience is one of the most precious memories of my life.

taoenwen:
Hi @crashGG 
I'm guessing you're referring to a cardless hack—achieving the functionality of having a card inserted without actually inserting one, particularly to unlock characters, items, etc. I'm not sure if my assumption is correct.

If that's the case, I suggest adding a switch (default off) in the DipSW, such as "Fake card?", and enabling this feature when the switch is turned on.

barbudreadmon:
@crashGG 
I have a feeling this PR is meant as a workaround for the libretro port which is still missing a mecanism for inserting those cards, you should have made a PR for that instead.

Otherwise, i don't really understand the usefulness of this PR, my understanding is that you can already achieve the same result by inserting a card.

dinkc64:
@crashGG ,
Are you referring to Gemini Wing by Tecmo?  This one has a nice weapons system - 
power weapons collected on the ships tail.   Though I played this one first on the C64
home computer, but now I only play the arcade version.  

best regards,
- dink

barbudreadmon:
Sadly there were no arcade room where i lived as a kid, afaik the closest arcade cabinets were in the neighbor town's pubs and i rarely had the chance to go there before high school. I was mostly playing consoles/computers at home/relatives/friends (odyssey², atari 2600, thomson MO5/TO7, atari st, nes, sms).
The only cabinet i remember playing "as a kid" is Ghosts'n Goblins, more specifically an earlier revision with the boss scroll bug :P.

crashGG:
> @crashGG I have a feeling this PR is meant as a workaround for the libretro port which is still missing a mecanism for inserting those cards, you should have made a PR for that instead.
> 
> Otherwise, i don't really understand the usefulness of this PR, my understanding is that you can already achieve the same result by inserting a card.



> @crashGG I have a feeling this PR is meant as a workaround for the libretro port which is still missing a mecanism for inserting those cards, you should have made a PR for that instead.
> 
> Otherwise, i don't really understand the usefulness of this PR, my understanding is that you can already achieve the same result by inserting a card.

Yes, inserting a card can achieve the same effect. However, this means that after each game starts, at least five mouse clicks (navigation + selection) are required on the UI to load the memory card and begin normal gameplay. Sometimes, a misclick—such as accidentally selecting "Create New Memory Card"—can overwrite the old default memory card. Additionally, this is not user-friendly for players who prefer launching the game in full-screen immersion mode. This PR allows players to forget this cumbersome "fourth-wall-breaking" operation and focus on the game itself.

crashGG:
> @crashGG , Are you referring to Gemini Wing by Tecmo? This one has a nice weapons system - power weapons collected on the ships tail. Though I played this one first on the C64 home computer, but now I only play the arcade version.
> 
> best regards,
> 
> * dink

Yes ,exactly,There were also a lot of vertical scrolling shooters at the same time,like twin cobra. Two years later, my interest shifted to Final Fight.

barbudreadmon:
>Yes, inserting a card can achieve the same effect. However, this means that after each game starts, at least five mouse clicks (navigation + selection) are required on the UI to load the memory card and begin normal gameplay. Sometimes, a misclick—such as accidentally selecting "Create New Memory Card"—can overwrite the old default memory card. Additionally, this is not user-friendly for players who prefer launching the game in full-screen immersion mode. This PR allows players to forget this cumbersome "fourth-wall-breaking" operation and focus on the game itself.

Forgive the ignorance, i haven't tested this yet in standalone. Is it something you need to do at every FBNeo standalone launch ? It doesn't seem that bad if you only need to do it once.

crashGG:
Why not give it a try? )

barbudreadmon:
>Why not give it a try? )

Because i'm not a windows user, so could you answer ?

crashGG:
> > Why not give it a try? )
> 
> Because i'm not a windows user, so could you answer ?

If you want to choose a card-specific character, yes. You'll need to perform the card insertion process described above every time you enter the game.

barbudreadmon:
@dinkc64 @taoenwen shouldn't memory card settings be saved in some .ini variable ? it might indeed be cumbersome if the user has to set everything again every time.

crashGG:
> @dinkc64 @taoenwen shouldn't memory card settings be saved in some .ini variable ? it might indeed be cumbersome if the user has to set everything again every time.

From a product design perspective, this won't work. Settings are based on each arcade cabinet(per game). If you reserve a memory card setting for a specific cabinet (game), it's like leaving a single personal memory card for every user on an arcade. Other users will complain: "I don't want to start the game with the previous user's configuration and character." Damn,  you can't even choose a new character before the game starts.

barbudreadmon:
Let alone the fact you can't use FBNeo in a public environment, hence the current user should be the same as the previous user most of the time. I don't really see what would be the problem here since the new user would just have to load his own memory card, just like he is forced to do currently.

From a product design perspective, your PR won't work because it prevents the game from being started with no character unlocked.

crashGG:
> Let alone the fact you can't use FBNeo in a public environment, hence the current user should be the same as the previous user most of the time. I don't really see what would be the problem here since the new user would just have to load his own memory card, just like he is forced to do currently.
> 
> From a product design perspective, your PR won't work because it prevents the game from being started with no character unlocked.

In other words, it removed the official restrictions and penalties imposed on players who did not insert cards.
Well, let more people judge. If feel that using this pr is inappropriate for the master line, I can move it to the Libretro port.

taoenwen:
While the no-card hack offers its own convenience, we should also take into account the experience of all players. After all, enabling hidden options from the very beginning of the game may not necessarily be a good thing:

[1] Some advanced players prefer a clean start, aiming to gain the greatest sense of achievement through their own efforts.

[2] Other casual players enjoy starting fully equipped—they may lack time or experience, but still don’t want to miss out on any of the excitement.

[3] There are also players who value the complete experience of the gameplay process​ (including inserting and ejecting the card), recreating the feeling of being in an arcade, much like the sensation of inserting a coin each time.

[4] And even more types of players exist…

I suggest implementing a DipSW toggle option, allowing users to decide whether to enable the feature or not. After all, the choice made by the users themselves is what feels most right to them at that moment.

dinkc64:
@taoenwen I also think the DIP option is the best idea for this kinda feature, then
everyone is happy

taoenwen:
> @taoenwen I also think the DIP option is the best idea for this kinda feature, then everyone is happy

Thank you, sir!

taoenwen:
Cardless mode

[off]
image

[on]
image

@crashGG

PR #2537, opened 2026-05-01
d_coleco.cpp: various...

fprietog:
Added the support to:
"BUSTin-Out Vol. 0 (HB)"
"Mr. Do! Arcade (HB)"

Fixed typo in "Schlange CV (HB)"


taoenwen:
Thanks!

PR #2545, opened 2026-05-02
d_megadrive.cpp: updated "Ghost Baseball (HB, v2.0)"

fprietog:
Version 2.0 is Live!

The Home Run Line is real.

There's now a finish line at the top of the score tunnel. Reach it and your at-bat ends in a celebration: an explosion, a HOME RUN banner, and a bonus payout. The bonus scales with how many heart rings you collected on the way up  so a clean run rewards both **distance** and **pickups**, not just one.

The line sits at altitude 19,296 (the 100th distance marker, labelled "1000"). Hard but reachable. Stars start streaking as you approach so you'll feel it coming.

Cheerleaders

A row of pixel cheerleaders waits at the home run line, doing a Mexican-wave bob along the corridor. Hi.

More patterns to dodge

Thirteen new ring-of-fielder formations: caret, snake, comb, two-gap, hourglass, zigzag, wedge, crescent, t-bone, bracket, cascade, and crossfire. Showpieces like Cascade (three staggered walls in a row) and Crossfire (alternating thrusts from each wall) only show up deeper in the tunnel

No more back-to-back repeats

The formation picker now remembers its last three picks and avoids repeating itself. You'll see more variety, especially on long runs.

Have fun. Let us know in the comments what your top score is — the seeds top out at 30,000, but the ceiling is much higher if you can chain rings.

Mike & Charlie  from Infinite State Games

### DIFF
--- a/src/burn/drv/megadrive/d_megadrive.cpp
+++ b/src/burn/drv/megadrive/d_megadrive.cpp
@@ -40035,10 +40035,10 @@ struct BurnDriver BurnDrvmd_genmtetris = {
 	&bMegadriveRecalcPalette, 0x100, 320, 224, 4, 3
 };
 
-// Ghost Baseball (HB, v1.3)
+// Ghost Baseball (HB, v2.0)
 // https://infinitestategames.itch.io/ghost-baseball
 static struct BurnRomInfo md_ghbaseballRomDesc[] = {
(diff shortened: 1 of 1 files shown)


fprietog:
@dinkc64 Now you have your HOME-RUN line!!!

dinkc64:
damn, I love these guys.  :)

PR #2560, opened 2026-05-09
adj. in d_sms.cpp / d_snes.cpp ...

KlaviculaAG:
add 'gg_frayte' english translation
add 'snes_fzeroml' hack.
adj. in lorettate (as on gg/sms roster).


taoenwen:
Thanks!

Issue #2561, opened 2026-05-09
orleg2_104cn crashed


lrf739146825:
<img width="598" height="499" alt="Image" src="https://github.com/user-attachments/assets/cddadd6e-3756-433b-aaf9-b7fa6a0dc553" />

[Savestates near Crash and zzBurnDebug.zip](https://github.com/user-attachments/files/27556504/Savestates.near.Crash.and.zzBurnDebug.zip)

dyoalie:
I've met the same problem earlier the day before (except for the role I played)

dinkc64:
Hi @lrf739146825  &  @dyoalie 
To better help us deal with this issue,
Would you please make an inputs recording and attempt to reproduce this bug?

After starting game, go to Game -> Replay... -> Record Input,
and make sure "[x] From Power-On" box is checked !
When the bug happens, go "Game -> Replay... -> Stop replay/record" and post the file
here (zip it first).

@hogenwang,
Hi,
Do you have any ideas about a possible solution for this issue?  
What do you think about using the newer Arm7+9es core?  I have a feeling it could be
due a cpu core bug (arm7 core in fbneo is very, very old).  But... it could also be a driver bug.

best regards,
- dink

lrf739146825:
[orleg2_104cn Replay.zip](https://github.com/user-attachments/files/27582301/orleg2_104cn.Replay.zip)
The crash occurs at the same phase of the same BOSS.

hogenwang:
Okay, I'll see how to fix this bug.

hogenwang:
Hi @lrf739146825 @dinkc64 @dyoalie 
I have fixed this issue and compiled a new version. Please test it together; if there are no problems, I will then submit these changes.
[domload fbneo](https://fbn.us3.ip53.com/fbneo64.rar)

best regards,

hogen

lrf739146825:
@hogenwang  Thank you for your fix. I have tested the client you released, the game runs well, and the previously occurring crashes have been resolved.

<img width="880" height="711" alt="Image" src="https://github.com/user-attachments/assets/a454f996-1512-491f-bcdd-fa80af3f7e61" />

dinkc64:
@hogenwang, seems to be a-ok ! :)

PR #2586, opened 2026-05-20
Fix orleg2_104cn crashed #2561

hogenwang:
Fix orleg2_104cn crashed #2561


dinkc64:
Hi @hogenwang
Thanks for the fix!

best regards,
- dink

PR #2596, opened 2026-05-22
d_outrun.cpp: added "Out Run (sitdown/upright, Rev B) (added music)"

fprietog:
This is an Out Run version that adds two new music tracks.

In music selection if you have the gear in "Low" position it show the usual three music tracks. But with gear in "High" position it shows at the left "CAMINO A MI AMOR" and at the right "CRUISING LINE".

ATM only the additional "CAMINO A MI AMOR" can be played because if you choose "CRUISING LINE" it plays "CAMINO A MI AMOR" too.


taoenwen:
Thanks!

PR #2626, opened 2026-06-12
d_megadrive.cpp - adj. in cvaniacotm

KlaviculaAG:
':' not allowed in filename. reported by ROM manager.


fprietog:
Thanks for the fix.

PR #2633, opened 2026-06-22
Minor misspelling fix in vid_switchres.cpp

MCaG12:


dinkc64:
Thanks for you contribution

best regards,
- dink

PR #2645, opened 2026-07-06
fixing warnings about unused variables in d_galaxian.cpp ,arm9.cpp and neocdsel.cpp

MCaG12:


dinkc64:
Hi @MCaG12 
Regarding arm9, those unused variables are used for debugging, and they might be needed as the pgm2 driver is relatively new, it would probably be a good idea to restore those for the time being

best regards,
- dink

MCaG12:
The debugging variables in arm9 have been readded

dinkc64:
thanks :)

PR #2647, opened 2026-07-07
some changes in d_neogeo, d_snes

Igor-Arabe:
updated zintrckbp - with both sound effects and audio tracks crashband en/pt updated to v1.23


taoenwen:
Thanks!

PR #2688, opened 2026-08-08
New Ghosts'n Goblins hacks: Enhanced Edition and Artoria Hacks + Enhanced Editions

GoodLuckTrying:


barbudreadmon:
If you want to apply an ips to a romset, FBNeo has a feature for this.
Please don't duplicate all romsets from a driver by applying that ips to all of them.
With that kind of incremental logic, this commit is 20 romsets, next will be 40.

PR #2738, opened 2026-08-31
Add GBA PPU worker thread component.

lrf739146825:
Offloads PPU rendering to a dedicated worker thread — ideal for CPUs with limited single-core performance but available multi-core headroom — delivering tangible framerate gains through parallel execution.

Use `#define PPU_WORKER_DISABLE` to disable the worker thread, which is useful for debugging non-thread issues on the GBA.

### DIFF
--- a/src/burn/drv/gba/apu.h
+++ b/src/burn/drv/gba/apu.h
@@ -18,7 +18,7 @@ static inline float gba_polyblep(float t, float dt)
 {
 	if (t <= dt) {
 		t = t / dt;
-		return t + t - t * t - 1.0;;
+		return t + t - t * t - 1.0f;
 	} else if (t >= 1 - dt) {
 		t = (t - 1.0) / dt;
 		return t * t + t + t + 1.0;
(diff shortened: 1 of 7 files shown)


barbudreadmon:
FBNeo has a tested cross-platform threading component at https://github.com/finalburnneo/FBNeo/blob/master/src/burn/devices/thready.h and ideally you should reuse it.

dinkc64:
I'm not sure what kinda thread they're dealing with here, but I should point out:

thready has its limitation - it basically spins up a single-task thread that's meant to do something cocurrently and then exit.
ie: it's not for threads that need to stay alive all the time and keep receiving/sending messages

lrf739146825:
Thanks for your guidance, I'll try to improve compatibility.This is a feature I completed with AI assistance, and I ran a lot of local simulation tests. The results show that multithreading is indeed great, but it's much more complicated to implement than single-threading, and doing it purely manually is quite discouraging.

barbudreadmon:
The ppu worker seems to only do something between the beginning (ppu_worker_begin_frame ?) and rendering (ppu_worker_vblank_publish ?) of frame, so thready is probably suitable ?

taoenwen:
This is a clever design. However, scanline‑level snapshots inherently force an automatic fallback away from per‑pixel rendering. If this path is to be retained, I suggest tying it explicitly to the scanline‑rendering dip switch. Otherwise, HDMA‑dynamically‑updated VRAM, OAM and palette effects will be lost.

The worker thread allocates a full `gba_t` structure on the stack for rendering, which introduces memory‑copy and initialization overhead. Real‑world testing is needed to verify whether this actually delivers improved frame rates on low‑end Android hardware.

There is also an unrelated change within this pull request that alters DMA‑prefetch timing; this should be tested against timing‑sensitive FC‑Mini titles.

Two self‑contained changes from this PR can be cherry‑picked independently:
- Skip pixel rendering for run‑ahead frames.
- Maintain updates to affine reference points even for non‑rendered frames.

lrf739146825:
This is not a good approach; it complicates the problem.

PR #2739, opened 2026-08-31
d_cps2: add mmatrixi

Igor-Arabe:
improvement hack [usa] + 3button inputList
for jap region use romdata/ips option.


dinkc64:
Thank you Igor-Arabe!

Issue #1718, opened 2024-03-13
Jurassic Park: transparency gitch on the first stage


vanfanel:
Hello there,

Running latest FBNeo GIT code here.
I've noticed a transparency glitch on the first stage of Jurassic Park, where brown-ish transparent bands move up and down with the cursor.
I recorded a video of it:

https://github.com/finalburnneo/FBNeo/assets/837585/5fd2ddd1-9717-4ef6-a08b-0b05cf5ad6bf

Seems like a Sega System 32 emulation bug.

dinkc64:
Hi vanfanel,
Here's how it's supposed to look: 
https://youtu.be/Mx0fHZOThlo?si=pflfqBvwG1VW1Y6R&t=153

I'll try to figure it out, but I can't promise anything, the Sega System 32 is kinda complex!

best regards,
- dink

mahoneyt944:
I believe it's related to row scroll. I managed to get it working for player 2 but not player 1 in mame2003+. Player 1 might just be a priority issue though as its not shown at all but never got any further with it. But maybe this will help here.

https://github.com/libretro/mame2003-plus-libretro/blob/3206819c48567e759f993134ac27491b11b3dd80/src/vidhrdw/system32_vidhrdw.c#L1330
![jpark-240613-021406](https://github.com/finalburnneo/FBNeo/assets/49591133/51522315-44d7-4c0a-8e8e-fea8654b69c8)

dinkc64:
Hi mahoneyt944, thanks for the tip!
I'll see if this can be improved upon,

best regards,
- dink

dinkc64:
@mahoneyt944 Odd, we already have those lines in a similar spot, was there anything else involved in the fix?

mahoneyt944:
@dinkc64 I went down the rabbit hole a little by porting newer segas32 video code over to mame2003+.... I get the same issue noted here. Curiously it's not broken for player 2 in any code base, so I'm hopeful that it's a minor adjustment needed for player 1. This issue has been present since the beginning of this video code (around mame 97u2), so it's unlikely my previous comment will help here. Player 2's render might point us in the right direction though.

mahoneyt944:
So this effect is created by `update_tilemap_rowscroll(&layer_data[MIXER_LAYER_NBG2], cliprect, 2);` during `update_tilemaps()` and seemingly works for player 2 as intended but not for player 1. I would think this is using row select and perhaps it's selecting the wrong row for player 1.

[...]

PR #2743, opened 2026-09-03
Port segas32 per line clipping - jpark, ga2, titlef, sonic ending

mahoneyt944:
Noticed the secty addition, has 6 conditions where it's used...but above where the clips are set only has 4 ifs in place of sect_rect. Might want to double check and make it all use the same. Can close out Segas32 related issue tickets....finally!

Mame2003-plus defines it like this
```
static INLINE void sect_rect(struct rectangle *dst, const struct rectangle *src)
{
	if (src->min_x > dst->min_x) dst->min_x = src->min_x;
	if (src->max_x < dst->max_x) dst->max_x = src->max_x;
	if (src->min_y > dst->min_y) dst->min_y = src->min_y;
	if (src->max_y < dst->max_y) dst->max_y = src->max_y;
}
```

### DIFF
--- a/src/burn/drv/sega/d_segas32.cpp
+++ b/src/burn/drv/sega/d_segas32.cpp
@@ -79,7 +79,7 @@ static UINT8 transparent_check[32][256];
 struct extents_list
 {
 	UINT8	scan_extent[256];
-	UINT16	extent[32][16];
+	UINT16	extent[256][16];
 };
 
 static UINT16 mixer_control[2][0x40];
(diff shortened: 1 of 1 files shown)


mahoneyt944:
#1718

dinkc64:
Thanks buddy!!  Awesome work! :)

I just pasted the code from another driver,  I'll probably make it a global function
to keep from needing to paste it again in the future

best regards,
- dink

